PR #5, opened 2017-07-12
Build from source

vorlonofportland:
Fix up our gadget snap to always build from the current .debs in Ubuntu instead of shipping opaque binaries in the repository.


cyphermox:
Looks file to me; if only that I wonder if we could also drop loopback and/or chain. That however would have to be for a different merge anyway.

PR #6, opened 2017-10-20
Stage grub packages needed for build instead of fetching as build-deps

sil2100:
Declaring the grub-related packages as build-dependencies in snapcraft.yaml can cause conflicts with packages installed on the host, like in the case where the host machine has grub-pc installed on a xenial system. This causes conflicts in satisfying build dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ GRUB_MODULES = \
 	video
 
 all:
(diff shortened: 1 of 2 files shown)


sil2100:
Please note that I am still new to snapcraft.yaml syntax. But I tested it for our purposes and it seems to generate the same gadget tree.

PR #10, opened 2018-12-26
Add mmx64.efi (MokManager) to support mokutil

tsunghanliu:
To support MokUtil properly, mmx64.efi (MokManger) should be included in the gadget snap.

Though there is no mokutil included in core and core18 and there is no mokutil snap, users still can upload the mokutil binary and run it by themselves. Another way is to include mokutil in core and core18.


tsunghanliu:
Hi @mvo5 and @zyga,
Could you please take a look these two PR #10 and #11, when you have time? Thanks.

PR #11, opened 2018-12-26
Add mmx64.efi (MokManager) to support mokutil

tsunghanliu:
Same thing for Ubuntu Core 18.

To support MokUtil, we need to include the mmx64.efi (MokManger).

Though there is no mokutil included in core and core18 and there is no mokutil snap, users still can upload the mokutil binary and run it by themselves.
Another approach is that we can include mokutil in core and core18 to simplify the operation.


vorlonofportland (approved):
+1

PR #14, opened 2019-05-20
gadget.yaml: add system-recovery partition

mvo5:
For UC20 we want to have a recovery partition. This adds to
the existing system-boot partition. The size is currently set
to 600Mb - this is a bit arbitrary and will allow us roughtly
2 recovery systems. We have the following sizes today:
 * core18:  54MB
 * snapd:   23MB
 * gadget:   1MB
 * kernel: 200MB
The recovery partition is setup after the system-boot
partition.

It also adds a initial grub.cfg which is very basic and
only supports a single recovery system at this point but
that recovery system can be selected via the grubenv:
* `snap_recovery_kernel`
* `snap_recovery_core`

Mostly a stub for now but it will allow us to start building a "pc" gadget in the "20/edge" channel.


gadget.yaml
```diff
@@ -26,3 +26,18 @@ volumes:
             target: EFI/boot/bootx64.efi
           - source: grub.cfg
             target: EFI/ubuntu/grub.cfg
+      - name: Recovery
+        filesystem: vfat
+        # FIXME: what type to use?
+        type: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
```
xnox: Hmmm.... That is ESP type, and you appear to have ESP like content too. I don't believe it is allowed as per UEFI spec to have multiple ESP partitions.

Recovery must be booted from the system-boot partition.... Or you need chainload from there....

mvo5: Thanks, I was suspecting this, I will change things.

recovery-grub.cfg
```diff
@@ -0,0 +1,28 @@
...
+menuentry "Normal mode" {
+          search --label "system-boot" --set=root
+          chainloader +1
+}
+
+set cmdline="root=LABEL=$label snap_recovery_mode=$snap_recovery_mode snap_core=core.snap snap_kernel=kernel.snap ro net.ifnames=0 console=ttyS0 console=tty1 panic=-1"
+
+for system in /system/*
+do
+  menuentry "Recovery $system" $system {
```
pedronis: why the multiple recovery entries, wasn't the plan do the choosing from initramfs?

recovery.md
```diff
@@ -0,0 +1,82 @@
...
+"writable" partition. The partitions are separate because we want to
+write to the recovery partition as rarely as possible to avoid
+filesystem issues.
+
+# Recovery system file layout
+
+The revery partition contains the following file layout:
+
+/system/<name>/snaps/{base.snap,kernel.snap,other...}
+/system/<name>/assertions.txt
```
pedronis: not a fan of the ".txt" extetnsion here, also unclear why we shouldn't support multiple streams

recovery.md
```diff
@@ -0,0 +1,82 @@
+# The recovery partition
+
+On Ubuntu Core 20 systems there is a new system-recovery partition. It
+is big enough to contain a certain number of recovery "systems" that
+can then be used to recover or reinstall a broken system.
+
+A full "seed" to reinstall on a generic amd64 system is roughly 280MB
```
pedronis, line 7: do we how big things like network manager etc are?  just 20mb for those seems tight?

gadget.yaml
```diff
@@ -14,8 +14,23 @@ volumes:
...
           - image: pc-core.img
-      - name: EFI System
+      - name: Recovery
+        filesystem: vfat
+        # the ESP partition
         type: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
+        # max volume label length: 11 chars
+        filesystem-label: sys-recover
+        # too big? too small?
+        size: 600M
```
zyga: There's a recommended size for windows systems that we should probably follow. I think it's around a GB now.

grub.cfg-recovery
```diff
@@ -0,0 +1,27 @@
...
+menuentry "Normal mode" {
+          search --label "system-boot" --set=root
+          chainloader ($root)/efi/boot/grubx64.efi
+}
+
+set cmdline="root=LABEL=$label snap_recovery_mode=$snap_recovery_mode snap_core=core.snap snap_kernel=kernel.snap ro net.ifnames=0 console=ttyS0 console=tty1 panic=-1"
+
+for system in /system/*
+do
+  menuentry "Recovery $system" $system {
```
mvo5: Remove this and just do have "recover", "install" - each will boot into a "select" boot mode which allows to select what recovery/install system to boot from the initramfs. After the selection a grubenv setting is used to do a one-off boot into this selected recovery/install system.

recovery.md
zyga: I think this should be a assertions.d style directory.

recovery.md
```diff
@@ -22,15 +22,18 @@ filesystem issues.
...
+Where <name> is an encoded date/time like 20190521-1213. The parition
+is FAT so we need to put the assertions in the "stream" format on disk
+with a short name. The assertion files must include exactly one model
+assertion.
 
 The names of the kernel and the base are fixed. This allows
-us a static grub.cfg menu.
+us a static grub.cfg menu. The selection of the recovery system will
+happen at a later time from initramfs. To do this the recovery system
+will boot into a sepecial "select" mode and then the selection is set
```
pedronis: special

recovery.md
```diff
@@ -44,7 +47,8 @@ account for now. This will change in a later revision of this doc.
 * always boot into system-recovery partition
 ** check if system is setup for normal booting
 *** if so, chainboot into the system-boot partition
-*** if not, boot into selected bootmode, set snap_recovery_mode to "recovery" or "install"
+*** if not, boot into recovery bootmode, set snap_mode="recovery"
+**** later the initamfs will allow selecting different recovery systems
```
pedronis: initramfs

recovery.md
```diff
@@ -22,15 +22,18 @@ filesystem issues.
...
 
 /system/<name>/snaps/{base.snap,kernel.snap,other...}
-/system/<name>/assertions.txt
+/system/<name>/assertions/<short-name>
 
-Where <name> is an encoded date/time like 20190521-1213. The
-parition is FAT so we need to put the assertions in the "stream"
-format on disk. The assertions.txt must include exactly one
-model assertion.
+Where <name> is an encoded date/time like 20190521-1213. The parition
```
pedronis: the partition

recovery.md
```diff
@@ -22,15 +22,18 @@ filesystem issues.
...
+is FAT so we need to put the assertions in the "stream" format on disk
+with a short name. The assertion files must include exactly one model
```
pedronis: with short filenames. The assertion files all together must ...

pedronis (commented):
just some comments on recovery.md. maybe it needs  a spellcheck pass

gadget.yaml
mvo5: I changed this to have a single ESP partition now - I think this makes it easier and more reliable. However in my reading I could not find anything about a limit of ESP partitions. I don't know if the boot order is defined if there are multiple ones so I think its wise to limit ourselfs to just one (especially since we don't need multiple ones).

recovery.md
mvo5: I increased this now.

gadget.yaml
mvo5: Yeah, Samuele also commented on this. I increased the size for now.

recovery-grub.cfg
mvo5: Yes, I changed it to a single static entry.

recovery.md
```diff
@@ -0,0 +1,82 @@
...
+unpacked kernel(s) to boot init initramfs that then unencrypts the
+"writable" partition. The partitions are separate because we want to
+write to the recovery partition as rarely as possible to avoid
+filesystem issues.
+
+# Recovery system file layout
+
+The revery partition contains the following file layout:
+
+/system/<name>/snaps/{base.snap,kernel.snap,other...}
```
mvo5, line 24: <name> should include the branch/model (directories) - we need to consider upper/lower. What to boot initially needs to be written by `snap prepare-image` in grubenv

pedronis: I can double check what are the relevant regexps in snapd

pedronis: * snap names should be like  `^[a-z0-9-]*[a-z][a-z0-9-]*$`  and at most 40 chars
* model names should be like `^[a-zA-Z0-9](?:-?[a-zA-Z0-9])*$` but we enforce lowercase on top of that, no size limits ATM afaict
* brand names  should be like `^(?:[a-z0-9A-Z]{32}|[-a-z0-9]{2,28})$`,  there sadly  uppercase can happen, same problem with snap-ids

mvo5: Uh, so in theory we might have "Brand" and "brand" and on vfat its the same - that is very unfortunate, I think that means we need to reconsider using branch/model in the directory.

pedronis: yes, but the example you gave is misleading,  nice brand names are enforced lowercase (like "canonical"), it's the random ones (same as snap ids) that can have uppercase, we can't assume the brand will be nice, is things like CA5GLZgNQWPhspDQK63Er46Uxz2SO7ez that have uppercase in them, it's unlikely they will collide but not impossible

mvo5:
This has not changed since a couple of days and we use it for our work to build images now. I would like to merge it into the *20* branch of the gadget and upload it to the "20/edge" channel for the "pc" snap in the store. But I need a :+1: review for this…

sil2100 (approved):
Ok, this looks generally good to me, but I don't know if the current recovery story changed from what I have heard in Capetown. I guess the sizes will be tweaked still, right?

Makefile
```diff
@@ -69,4 +69,4 @@ all:
 
 install:
 	install -m 644 pc-boot.img pc-core.img shim.efi.signed grubx64.efi $(DESTDIR)/
-	install -m 644 grub.conf grub.cfg $(DESTDIR)/
+	install -m 644 grub.cfg-recovery grub.cfg-normal grub.conf $(DESTDIR)/
```
vorlonofportland, line 72: 'grub.conf' is not a native name for the config file on grub's side, this name exists for compatibility with snapd's requirements.  If there are now two grub.cfg's, should the grub.conf symlink be dropped?

gadget.yaml
```diff
@@ -14,15 +14,31 @@ volumes:
...
         content:
           - image: pc-core.img
-      - name: EFI System
+      - name: Recovery
+        role: system-recovery
+        filesystem: vfat
+        # UEFI will boot the ESP partition by default first
         type: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
+        # max volume label length: 11 chars
+        filesystem-label: sys-recover
```
vorlonofportland: Per London sprint, name should now be 'Ubuntu-seed' instead?

gadget.yaml
```diff
@@ -14,15 +14,31 @@ volumes:
...
+          - source: grubx64.efi
+            target: EFI/boot/grubx64.efi
+          - source: shim.efi.signed
+            target: EFI/boot/bootx64.efi
+          - source: grub.cfg-recovery
+            target: EFI/ubuntu/grub.cfg
+      - name: System boot
+        type: 0B,EBD0A0A2-B9E5-4433-87C0-68B6B72699C7
         filesystem: vfat
         filesystem-label: system-boot
```
vorlonofportland: Per London: Ubuntu-boot

gadget.yaml
```diff
@@ -14,15 +14,31 @@ volumes:
...
-        size: 50M
+        size: 100M
```
vorlonofportland, line 35: I'm skeptical of this as a size for the partition given that we will need to store 2x vmlinuz (+ embedded initramfs) on this fs for the FDE case.

grub.cfg-normal
```diff
@@ -37,6 +37,6 @@ set cmdline="root=LABEL=$label snap_core=$snap_core snap_kernel=$snap_kernel ro
 menuentry "$snap_menuentry" {
     search --label $label --set=writable
     loopback loop ($writable)/system-data/var/lib/snapd/snaps/$snap_kernel
-    linux (loop)/kernel.img $cmdline
+    linux (loop)/kernel.img $cmdline systemd.debug-shell=1
```
vorlonofportland, line 40: This looks like an unrelated change that doesn't belong on the actual core20 branch.

grub.cfg-recovery
```diff
@@ -0,0 +1,40 @@
...
+insmod part_gpt
+
+if [ -s $prefix/grubenv ]; then
+  load_env
+fi
+
+# FIXME: find a better way? this ensures that someone instructed snapd
+#        to go into "install" mode it actually does that
+if [ "$snap_mode" = "install" ]; then
+    set default=1
```
vorlonofportland, line 13: Discussed in London that these don't need to be menu entries at all: if the chooser interaction is expected to be entirely via the initramfs, this could execute the boot commands inline instead of driving a menu selection.

vorlonofportland (changes requested):

xnox:
There are two branches in the repo one is called `20` the other is called `uc20` can we just use one or the other? Or make server-side gitbranch reference link to make one point to the other? (not sure if github supports that).

vorlonofportland:
> There are two branches in the repo one is called `20` the other is called `uc20` can we just use one or the other? Or make server-side gitbranch reference link to make one point to the other? (not sure if github supports that).

The uc20 branch is new, and appears to be a copy of 20.  I guess this is based on the discussion about wanting consistent naming of uc20 branches across repos, and we should consider uc20 authoritative now?  (Thus this PR will eventually need to be retargeted)

vorlonofportland:
In fact, this PR appears to be superseded by https://github.com/snapcore/pc-amd64-gadget/pull/19.

PR #15, opened 2019-05-22
add .travis.yml to enable basic smoke testing

mvo5:
This adds a .travis.yml that will build the snap to ensure
that this part works. With that in place we can add additional
tests later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: python
+os: linux
+dist: xenial
+sudo: enabled
+script:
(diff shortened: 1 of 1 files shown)


zyga (commented):
+1

Did you check that the way to use snaps follows the official travis guidelines?

mvo5:
@zyga Excellent point, I poked around and found that there is a new "addons" stanza in travis to support snaps. I use that now.

vorlonofportland (approved):

PR #19, opened 2019-07-05
UC20 spike changes

cmatsuoka:
This is the set of UC20 spike changes up to London names and extracted kernel for encrypted ubuntu-data.


grub.cfg-normal
```diff
@@ -31,12 +31,11 @@ elif [ "$snap_mode" = "trying" ]; then
...
+set label="ubuntu-data"
 set cmdline="root=LABEL=$label snap_core=$snap_core snap_kernel=$snap_kernel ro net.ifnames=0 init=/lib/systemd/systemd console=ttyS0 console=tty1 panic=-1"
 
 menuentry "$snap_menuentry" {
-    search --label $label --set=writable
-    loopback loop ($writable)/system-data/var/lib/snapd/snaps/$snap_kernel
-    linux (loop)/kernel.img $cmdline
-    initrd (loop)/initrd.img
+    search --label ubuntu-boot --set=ubuntu_boot
+    linux ($ubuntu_boot)/kernel.img $cmdline systemd.debug-shell=1
```
vorlonofportland: 'systemd.debug-shell=1' 
This looks like an unrelated change that doesn't belong on the actual core20 branch.

gadget.yaml
```diff
@@ -14,15 +14,29 @@ volumes:
...
+            target: EFI/boot/bootx64.efi
+          - source: grub.cfg-recovery
+            target: EFI/ubuntu/grub.cfg
+      - name: System boot
+        type: 0B,EBD0A0A2-B9E5-4433-87C0-68B6B72699C7
         filesystem: vfat
-        filesystem-label: system-boot
-        size: 50M
+        filesystem-label: ubuntu-boot
+        size: 100M
```
vorlonofportland, line 35: I'm skeptical of this as a size for the partition given that we will need to store 2x vmlinuz (+ embedded initramfs) on this fs for the FDE case.

Makefile
```diff
@@ -69,4 +69,4 @@ all:
 
 install:
 	install -m 644 pc-boot.img pc-core.img shim.efi.signed grubx64.efi $(DESTDIR)/
-	install -m 644 grub.conf grub.cfg $(DESTDIR)/
+	install -m 644 grub.cfg-recovery grub.cfg-normal grub.conf $(DESTDIR)/
```
vorlonofportland, line 72: 'grub.conf' is not a native name for the config file on grub's side, this name exists for compatibility with snapd's requirements. If there are now two grub.cfg's, should the grub.conf symlink be dropped?

grub.cfg-recovery
```diff
@@ -0,0 +1,40 @@
...
+insmod part_gpt
+
+if [ -s $prefix/grubenv ]; then
+  load_env
+fi
+
+# FIXME: find a better way? this ensures that someone instructed snapd
+#        to go into "install" mode it actually does that
+if [ "$snap_mode" = "install" ]; then
+    set default=1
```
vorlonofportland, line 13: Discussed in London that these don't need to be menu entries at all: if the chooser interaction is expected to be entirely via the initramfs, this could execute the boot commands inline instead of driving a menu selection.

vorlonofportland (changes requested):

grub.cfg-normal
mvo5: This is there to make debugging easier, if you feel strongly about it we could remove it. Note that this is the "uc20" spike branch we are merging into - this is not a "production" branch or anything like this.

gadget.yaml
mvo5: What size do you have in mind? I just checked and the kernel is right now around 8Mb for vmlinuz and around 7Mb for the initamfs (this includes some basic tools to unlock the LUKS device).

Makefile
mvo5: I think we can drop the symlink now - I'm not sure about the consequences though - I would like to do that in a followup so that if things break we don't need to look at too many variables.

grub.cfg-recovery
mvo5: Yes, mostly for convenience today to see things flipping from "recovery" -> "normal" (a bit of a debug aid).

Makefile
vorlonofportland: +1 to deferring.

gadget.yaml
vorlonofportland: initramfs will include all of the microcode updates, which are currently ~2.5M; one golang binary (snapd is currently 17M on amd64); systemd (~2M or so?); C runtime (at least 14M, based on the library dependencies of systemd+udev).

The microcode doesn't get compressed.  The other bits may get compressed, but there's an ongoing discussion whether compressing them negatively impacts boot speed, so we shouldn't assume they are compressed (or compressed much).

That's 35.5M for the initramfs, not counting any kernel modules which might need to be shipped in the embedded initramfs instead of built in to the kernel image; plus 8.4M for vmlinuz, which puts us at ~44M.

If we expect these systems to be usable for 10 years, we'll need to expect some expansion of kernels, if only due to the addition of drivers for new hardware.

So 150 or 200M seems more suitable to me (rounded to 4MiB).

grub.cfg-normal
vorlonofportland: How will we make sure we clean up all debugging-only code when we go to production?  Spike or not, that's what worries me since this doesn't interfere with the booting of the system but does leave a security hole around.

gadget.yaml
xnox: 30M	ubuntu-core-initramfs.img
9.0M /boot/vmlinuz-5.2.0-8-generic
is what i have currently, without firmware, microcode, snapd. But this has too many kernel modules in the initrd.

grub.cfg-normal
cmatsuoka: As agreed before starting the spike, this branch won't be merged into master -- it will serve only as a reference and all changes will be properly rewritten at a later moment, which should serve as a filter to remove debugging code. That said, most early problems related to snapd seeding have already been fixed so I believe we don't need the debug shell anymore. I'll push a change to remove it.

xnox:
This looks good, impossible to build with core20 based atm, will switch to core18 base for the time being.

PR #22, opened 2019-10-22
Store changes

xnox:
@cmatsuoka these are changes that are current in the snapstore.

However, I don't think this is what should exists in 20/edge channel now.

What is the latest gadget for 20, and can I merge it into 20 branch here now?

Ie. Is https://github.com/cmatsuoka/pc-amd64-gadget/tree/writable-ramdisk actually what we should be using now?

### DIFF
--- a/grub.cfg-normal
+++ b/grub.cfg-normal
@@ -1,5 +1,5 @@
 set default=0
-set timeout=0
+set timeout=3
 
 insmod part_gpt
 insmod ext2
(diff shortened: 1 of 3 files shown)


xnox:
Oh, but the 20 branch has everything that is in writable-ramdisk.

So I guess we simply need to build the 20 branch and upload it into the store.

mvo5 (approved):
Lets merge this and iterate

PR #23, opened 2019-11-07
Add missing partitions and improve grub.cfg-recovery

mvo5:
This PR works torwards a booting uc20 system. It adds the missing partitions to the gadget.yaml so that snap-bootstrap will be able to create them at boot. 

It also adds the grub.cfg-recovery work from John. We need to add code to snapd (snap prepare-image to be precise) next to ensure that the kernel/initrd are unpacked when the firstboot code runs. With that piece we should be able to reach the initramfs.

Adding a bunch of people as reviewers - this is just for awareness. Fwiw, I created an image with this and it seems to be working fine, i get the file in the right places, grub boots correctly. There is just no kernel extracted yet (which needs to happen as it has to be a signed kernel eventually).

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -14,7 +14,7 @@ volumes:
         offset-write: mbr+92
         content:
           - image: pc-core.img
-      - name: Recovery
+      - name: UbuntuSeed
         role: system-seed
         filesystem: vfat
         # UEFI will boot the ESP partition by default first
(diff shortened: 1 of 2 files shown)


gadget.yaml
```diff
@@ -27,3 +27,23 @@ volumes:
...
+        # whats the appropriate size?
+        size: 750M
+        content:
+          - source: grubx64.efi
+            target: EFI/boot/grubx64.efi
+          - source: shim.efi.signed
+            target: EFI/boot/bootx64.efi
+          - source: grub.cfg-normal
+            target: EFI/ubuntu/grub.cfg
+      # XXX: should we define the  "ubuntu-save" partition here too?
```
xnox, line 43: didn't we agree to put it at the end? I can't remember now.

gadget.yaml
```diff
@@ -27,3 +27,23 @@ volumes:
             target: EFI/boot/bootx64.efi
           - source: grub.cfg-recovery
             target: EFI/ubuntu/grub.cfg
+      - name: UbuntuBoot
+        role: system-boot
```
xnox, line 31: Do all roles have implicit default labels, or do we need to specify labels throughout?

xnox (commented):
Some comments / questions.

gadget.yaml
mvo5: They have implicit default labels for now

gadget.yaml
mvo5: We did think about putting ubuntu-save at the end but we don't have language today to say: "use all but N-bytes for ubuntu-data and then put the N-byte partition ubuntu-save at the end". I think we should not block on this for edge and revisit.

PR #24, opened 2019-11-18
gadget.yaml: do not use CamelCase for names

mvo5:
This is more in-line with what we discussed for the partition names I think.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -14,7 +14,7 @@ volumes:
         offset-write: mbr+92
         content:
           - image: pc-core.img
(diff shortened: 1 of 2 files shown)


xnox:
Yes, I also thought the right names were 'lower-case' not 'CamelCase', thought this was a deliberate change in naming to CamelCase.

xnox (approved):

mvo5:
@xnox Thanks for the review, it was not deliberate, just an accident I think. Unfortunately it seems travis/snapcraft is unhappy about building for core20 :/

PR #25, opened 2019-12-10
Fix kernel path in recovery grub.cfg

cmatsuoka:
The bootenv variable will now contain the full kernel path to allow
us to load the kernel from different locations, so remove the hardcoded
path from the configuration file.

Signed-off-by: Claudio Matsuoka <claudio.matsuoka@canonical.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,4 +69,4 @@ all:
 
 install:
 	install -m 644 pc-boot.img pc-core.img shim.efi.signed grubx64.efi $(DESTDIR)/
-	install -m 644 grub.cfg-recovery grub.cfg-normal grub.conf $(DESTDIR)/
+	install -m 644 grub.cfg-recovery grub.cfg-normal grub-recovery.conf grub.conf $(DESTDIR)/
--- a/grub-recovery.conf
+++ b/grub-recovery.conf
@@ -1,0 +1,1 @@
+grub.cfg-recovery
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -69,4 +69,4 @@ all:
 
 install:
 	install -m 644 pc-boot.img pc-core.img shim.efi.signed grubx64.efi $(DESTDIR)/
-	install -m 644 grub.cfg-recovery grub.cfg-normal grub.conf $(DESTDIR)/
+	install -m 644 grub.cfg-recovery grub.cfg-normal grub-recovery.conf grub.conf $(DESTDIR)/
```
xnox, line 72: I wonder if we really need two real files (one of which is unused), and two symlinks, both pointing at grub.cfg-recovery =)

I guess this is a todo to clean up. Originally grub.conf was meant to point at one or the other. Or something like that?!

xnox (approved):

xnox:
Currently I'm getting "error: cannot find boot config in "/tmp/tmpfs6qzwyd/unpack/gadget"" with the new snapd, and old gadget. Will merge this, push to store, and rebuild the image again to see if this will then work.

Makefile
cmatsuoka: I also wondered if the files could be renamed instead of using symlinks (or maybe the rest of the buildsystem expects the original names?). We can check this with @mvo5.

xnox:
WIth this in place, image builds.

xnox:
Released to 20/stable

PR #28, opened 2019-12-16
grub.cfg-normal: set snapd_recovery_mode= run in run mode

mvo5:
The grub.cfg-normal will run from the ubuntu-boot partition. In
order for "the-tool" to run and call snap-bootstrap to setup the
right mount in run-mode this kernel commandline has to be set.

The-tool has a "ConditionKernelCommandLine=snapd_recovery_mode"
so it only runs with that.

There is a bunch more cleanup that should be done in this file but to
keep churn minimal it just does this one change.

### DIFF
--- a/grub.cfg-normal
+++ b/grub.cfg-normal
@@ -32,7 +32,7 @@ elif [ "$snap_mode" = "trying" ]; then
 fi
 
 set label="ubuntu-data"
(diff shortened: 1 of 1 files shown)


xnox:
I'd rather drop the restriction on the-tool. Originally added it, because the-tool / snap-bootstrap at the time could not handle the run mode, but it can now.

xnox:
actually snap-bootstrap itself needs it.

PR #31, opened 2019-12-18
grub.cfg: switch to new style kernel_status failover handling

anonymouse64:
This implements the new UC20 spec for loading kernels and failover handling.

Note this is *blocked* until snapd sets the appropriate kernel_status and extracts the kernel assets to ubuntu-boot partition.

The one open question I had about this is what to do if `kernel_status` is not `"try"`, `"trying"`, or `""`. I implemented it such that we default to `""` if `kernel_status` is anything else.

I also renamed grub.cfg-normal to grub.cfg-boot because this grub lives on the ubuntu-boot partition and it's easier to understand which grub corresponds to which partition with this naming scheme I think.

How to test right now:

1. Build an uc20 image with ubuntu-image and a local build of this gadget snap
1. Mount the raw image built by ubuntu-image and add a new ubuntu-boot partition to it, make that a ext4 filesystem
1. Mount the ubuntu-boot partition locally and copy the `grub assets`, `grubx64.efi`, `bootx64.efi` and grub.cfg into `$ubuntu-boot/EFI/boot/grubx64.efi`, `$ubuntu-boot/EFI/boot/bootx64.efi` and `$ubuntu-boot/EFI/ubuntu/grub.cfg` respectively.
1. Generate a grubenv without any variables defined at `$ubuntu-boot/EFI/ubuntu/grubenv` to modify later at runtime
1. Copy a working kernel asset into the ubuntu-boot partition at `$ubuntu-boot/kernel.efi` from the kernel snap
1. Put a "broken" kernel in the ubuntu-boot partition at `$ubuntu-boot/try-kernel.efi` (I just put in a file with "broken" in it there)
1. Unmount the partition and boot the image with qemu (or similar):

```
$ kvm -m 2048 -netdev user,id=mynet0,hostfwd=tcp::8022-:22,hostfwd=tcp::8090-:80 -device virtio-net-pci,netdev=mynet0 -bios /usr/share/OVMF/OVMF_CODE.ms.fd -drive if=virtio,file=uc20-try-kernel.img,format=raw -serial mon:stdio -nographic
```

8. In the first ubuntu-seed grub, choose "Continue into run mode" option
8. Let ubuntu-boot grub boot the default kernel for long enough to see that it is generally working
8. Reboot, and choose "Continue into run mode" option again
8. Enter the grub command line and run the following to confirm that kernel_status is currently unset and then modify it to `"try"` and reboot:

```
grub> load_env --file /EFI/ubuntu/grubenv kernel_status
grub> echo $kernel_status

grub> set kernel_status=try
grub> save_env kernel_status
```
12. Choose "Continue into run mode" option in the ubuntu-seed grub
12. Observe that ubuntu-boot grub fails to boot and reboot the machine:

```
alloc magic is broken at 0x5c807840: 0
Aborted. Press any key to exit.
```
14. Choose "Continue into run mode" option in the ubuntu-seed grub
14. Enter the grub command line and echo `kernel_status` to see it has been reset to `""`:

```
grub> load_env --file /EFI/ubuntu/grubenv kernel_status
grub> echo $kernel_status

grub>
```


grub.cfg-boot
```diff
@@ -0,0 +1,38 @@
...
+    menuentry "Run Ubuntu Core 20" {
+        search --label ubuntu-boot --set=ubuntu_boot
+        chainloader ($ubuntu_boot)/try-kernel.efi $cmdline
+    }
+elif [ "$kernel_status" = "trying" ]; then
+    # nothing cleared the "trying snap" so the boot failed
+    # we clear the mode and boot normally
+    set kernel_status=""
+    save_env kernel_status
+elif [ x"$kernel_status" != x"" ]; then
```
cmatsuoka: Is the x technique needed in grub scripts? Or maybe we could simply test for non-emptiness using `-n`?

anonymouse64: Yes, I wasn't clear on that, I think we might be able to just use `-n`, let me check

grub.cfg-boot
```diff
@@ -0,0 +1,38 @@
+set default=0
+set timeout=3
+
+# load only kernel_status from the bootenv
+load_env --file /EFI/ubuntu/grubenv kernel_status
+
+set cmdline="console=ttyS0 console=tty1 panic=-1 systemd.gpt_auto=0 snapd_recovery_mode=run"
+
```
xnox, line 8:
```suggestion
set kernel=kernel.efi
```

grub.cfg-boot
```diff
@@ -0,0 +1,38 @@
...
+if [ "$kernel_status" = "try" ]; then
+    # a new kernel got installed
+    set kernel_status="trying"
+    save_env kernel_status
+
+    # try to boot with try-kernel.img
+    menuentry "Run Ubuntu Core 20" {
+        search --label ubuntu-boot --set=ubuntu_boot
+        chainloader ($ubuntu_boot)/try-kernel.efi $cmdline
```
xnox:
```suggestion
set kernel=try-kernel.efi
```

grub.cfg-boot
xnox: yes use `-n`

grub.cfg-boot
```diff
@@ -0,0 +1,38 @@
...
+elif [ x"$kernel_status" != x"" ]; then
+    # ERROR invalid kernel_status state, reset to empty
+    echo "invalid kernel_status!!!"
+    echo "resetting to empty"
+    set kernel_status=""
+    save_env kernel_status
+fi
+
+# boot with the normal kernel.img and initrd.img
+if [ x"$kernel_status" = x"" ]; then
```
xnox:
```suggestion
```

grub.cfg-boot
```diff
@@ -0,0 +1,38 @@
...
+    menuentry "Run Ubuntu Core 20" {
+        search --label ubuntu-boot --set=ubuntu_boot
+        chainloader ($ubuntu_boot)/kernel.efi $cmdline
```
xnox:
```suggestion
        chainloader ($ubuntu_boot)/$kernel $cmdline
```

xnox (commented):
Given that the menuentry should always be disabled, simply toggle the kernel name one is booting, and keep just the single menuentry at the end with dynamic kernel name (either stock, or try one).

I like the grub-boot name. However, I thought we must name the grub.cfg something particular in the gadget. I wish we'd stop shipping 4 grub configs, and limit ourselves to just two with like unique names.

anonymouse64:
@xnox done, applied your suggested changes and rebased on 20.

anonymouse64:
Snapd will be ready for this change with https://github.com/snapcore/snapd/pull/7947, so this PR and that one should happen at the same time I think

xnox:
@anonymouse64 with a small tweak we can de-couple the two.

One can test for presence of the unpacked kernel, if there isn't one, fallback to the "just boot anything one can find on ubuntu-data" which is what we currently have.

PR #32, opened 2019-12-19
gadget.yaml: increase default size of ubuntu-data to 3G

mvo5:
Until we have auto-grow of the partition or auto-size from the
snap-bootstrap create-partitions code we need a bigger default
so that we can actually run spread tests on this.


xnox (approved):

xnox:
Does this go to 12? =)

PR #34, opened 2020-02-12
Makefile: add "regex" to the modules for the non-uefi grub

mvo5:
This should allow booting of UC20 systems (without encryption) on
non-UEFI systems.


xnox (approved):

zyga (commented):
*cough* ;)

PR #37, opened 2020-02-24
Drop duplicate grub.cfg-*

xnox:
Only grub.conf and grub-recovery.conf are used by snapd, drop useless symlink indirection which only clutter the snap contents.


mvo5 (approved):
Thank you

mvo5:
This appears to have conflicts

PR #38, opened 2020-02-28
grub.cfg-recovery: fix typo, filter vars when loading them

anonymouse64:
This had a typo of `snapd_recovery` instead of `snapd_recovery_system`.

Also, as per the spec, we should only be loading specific vars from the grubenv's, which AIUI is more of a protection against programming errors than it is a physical attack vector.

### DIFF
--- a/grub.cfg-recovery
+++ b/grub.cfg-recovery
@@ -1,7 +1,7 @@
 set default=0
 set timeout=3
 
-load_env --file /EFI/ubuntu/grubenv
+load_env --file /EFI/ubuntu/grubenv snapd_recovery_mode snapd_recovery_system
 
 # if no default boot mode set, pick one
 if [ -z "$snapd_recovery_mode" ]; then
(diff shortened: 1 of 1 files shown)


anonymouse64:
For whatever reason, travis is happy and green, but that hasn't propagated to GitHub :confused:

grub.cfg-recovery
```diff
@@ -1,7 +1,7 @@
 set default=0
 set timeout=3
 
-load_env --file /EFI/ubuntu/grubenv
+load_env --file /EFI/ubuntu/grubenv snapd_recovery_mode snapd_recovery_system
```
bboozzoo, line 4: Won't this make it too hard to tweak or override some grub config?

bboozzoo: Nvm, looked at the specs and filtering is what we want.

bboozzoo (approved):

xnox (approved):

PR #39, opened 2020-03-30
Drop no longer used snap_kernel fallback

xnox:
I think this is a historic piece we did whilst bootstrapping uc20 and is no longer needed.


anonymouse64:
+1

The PR description / commit message don't match the code, but yes this code looks correct. For now, we will only want to load only the `snapd_recovery_kernel` boot variable from the recovery systems grubenv.

anonymouse64 (approved):
(actually github style approving)

PR #40, opened 2020-04-05
gadget.yaml: reduce data partition size to 1G

cmatsuoka:
The data partition is now extended during install to use all the
space available in the block device, so we can set the default
size back to a lower value to generate smaller image files.

Signed-off-by: Claudio Matsuoka <claudio.matsuoka@canonical.com>

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -45,5 +45,4 @@ volumes:
         role: system-data
         filesystem: ext4
         type: 83,0FC63DAF-8483-4772-8E79-3D69D8477DE4
-        # XXX: make auto-grow to partition
-        size: 3G
+        size: 1G


anonymouse64 (approved):
LGTM

(note the change in snapd which did this is https://github.com/snapcore/snapd/pull/8253)

xnox:
what size images should be produced on cdimage.ubuntu.com? I still default those to like 10G, what should it be there? something very small? or something with a little bit of spare size?

vorlonofportland:
The images targeted to VMs specify a sensible default size since whatever size we declare there is the size of the disk qemu will give you.  The images targeted to hardware should still be minimized, and expanded on first boot.

xnox (approved):

PR #44, opened 2020-04-24
grub-recovery.conf: don't make run mode default just because ubuntu-boot exists

anonymouse64:
For the case of re-installing or recovering, we will still have the ubuntu-boot partition around, but we explicitly don't want to make run mode the default here because we will have snapd_recovery_system set and snapd_recovery_mode set too, and so grub should use that.

### DIFF
--- a/grub-recovery.conf
+++ b/grub-recovery.conf
@@ -23,7 +23,6 @@ fi
 search --no-floppy --set=boot_fs --label ubuntu-boot
 
 if [ -n "$boot_fs" ]; then
-    default="run"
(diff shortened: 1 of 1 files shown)


xnox:
Ok this is buggy. But then there is inverse issue.

snap prepare image does not create grubenv file for the very first boot. And we default to install mode. If grubenv is corrupted or non existent, I was using presence of boot partition to still go into run mode instead of "wiping all of your data" mode.

xnox (approved):

PR #51, opened 2020-10-19
gadget: add ubuntu-save

bboozzoo:
Add new ubuntu-save partition, with role system-save. Note that snapd 2.48+ is
needed to work with this gadget.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -45,7 +45,11 @@ volumes:
             target: EFI/boot/grubx64.efi
           - source: shim.efi.signed
             target: EFI/boot/bootx64.efi
-      # XXX: should we define the  "ubuntu-save" partition here too?
+      - name: ubuntu-save
+        role: system-save
+        filesystem: ext4
+        type: 83,0FC63DAF-8483-4772-8E79-3D69D8477DE4
+        size: 8M
       - name: ubuntu-data
         role: system-data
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you!

gadget.yaml
```diff
@@ -45,7 +45,11 @@ volumes:
             target: EFI/boot/grubx64.efi
           - source: shim.efi.signed
             target: EFI/boot/bootx64.efi
-      # XXX: should we define the  "ubuntu-save" partition here too?
+      - name: ubuntu-save
+        role: system-save
+        filesystem: ext4
+        type: 83,0FC63DAF-8483-4772-8E79-3D69D8477DE4
+        size: 8M
```
bboozzoo: I ran into some issues with formatting this device as LUKS2 and then mounting it. We may need to use a larger size.

xnox: I don't think 8MB is enough for all the luks headers & backup, as well as ext4.

Is the intention for it to have 8MB of usable free disk space, once it's encrypted and mounted?

xnox:
Since images are built in livecd-rootfs livefs builders using snapd from deb; 2.48 needs to be in focal-updates before this gets merged.

bboozzoo:
@xnox could we land the PR and publish the snap in edge only?

PR #52, opened 2021-02-26
Implement shim fallback protocol

xnox:
With this change one can build gadget everywhere. As the required UC20 signed artefacts are pulled directly, rather than assumed to be available on the build host.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,11 +69,17 @@ all:
 	# And yet people try to rebuild this gadget snap
 	pull-lp-debs -a amd64 -D ppa --ppa ppa:canonical-foundations/uc20-staging-ppa shim-signed focal
 	dpkg-deb -x shim-signed_*.deb shim/
+	pull-lp-debs -a amd64 -D ppa --ppa ppa:canonical-foundations/uc20-staging-ppa shim focal
+	dpkg-deb -x shim_*.deb shim/
 	pull-lp-debs -a amd64 -D ppa --ppa ppa:canonical-foundations/uc20-staging-ppa grub-efi-amd64-signed focal || wget https://launchpad.net/~canonical-signing/+archive/ubuntu/uc20/+build/19903679/+files/grub-efi-amd64-signed_1.142.5+uc20.1+2.04-1ubuntu26.3_amd64.deb
 	dpkg-deb -x grub-efi-amd64-signed_*.deb grub/
 	cp shim/usr/lib/shim/shimx64.efi.dualsigned shim.efi.signed
+	cp shim/usr/lib/shim/fbx64.efi .
+	cp shim/usr/lib/shim/mmx64.efi .
+	cp shim/usr/lib/shim/BOOTX64.CSV .
 	cp grub/usr/lib/grub/x86_64-efi-signed/grubx64.efi.signed grubx64.efi
 
 install:
 	install -m 644 pc-boot.img pc-core.img shim.efi.signed grubx64.efi $(DESTDIR)/
+	install -m 644 fbx64.efi mmx64.efi BOOTX64.CSV $(DESTDIR)/
 	install -m 644 grub.conf $(DESTDIR)/
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,15 @@ all:
 	# BIOS boot partition must be defined with an absolute offset.  The
 	# particular value here is 2049, or 0x01 0x08 0x00 0x00 in little-endian.
 	/bin/echo -n -e '\x01\x08\x00\x00' | dd of=pc-core.img seek=500 bs=1 conv=notrunc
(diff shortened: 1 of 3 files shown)


xnox:
@superm1 so i'm finally getting around to adding proper shim fallback support. Such that on first boot ubuntu bootorder entry is populated and set as is typical. Because doing removable media path only does not work for a few usecases which have now been identified.

Ideally, I would make this compatible with the fwupd snap out of the box. As i do not want fwup to create boot entries, or to change bootorder. It can set BootNext and clear that only.

Where is the code for that? I.e. ideally i want to sign vendor and ship fwupd.efi with the UC20 signing key in the gadget and add the appropriate bootmenu entry for it in the BOOTX64.CSV file.

superm1:
@xnox the code to install and set all that up is here: https://github.com/fwupd/fwupd/blob/master/plugins/uefi-capsule/fu-uefi-bootmgr.c#L298

>Where is the code for that? I.e. ideally i want to sign vendor and ship fwupd.efi with the UC20 signing key in the gadget and add the appropriate bootmenu entry for it in the BOOTX64.CSV file.

So the potential challenge is if the "interface" between the EFI file ever changes then you suddenly can't use the snap unless the gadget is updated in tandem.  It's defined here: https://github.com/fwupd/fwupd/blob/master/plugins/uefi-capsule/efi/fwup-efi.h#L12
It hasn't changed since we subsumed fwupdate into fwupd 3 years ago, but I can't predict if the future we find a need to update it again.

bugraaydogar:
@xnox when should we expect the work to be completed? It is not possible to build a custom gadget within the current status of the repository. Currently, I will use the release version `20.20200210`

xnox:
> when should we expect the work to be completed?

I was going to merge this when we get back new shim, which is awaiting external action.

> It is not possible to build a custom gadget within the current status of the repository.

Yes it should be....

xnox:
> @xnox when should we expect the work to be completed? It is not possible to build a custom gadget within the current status of the repository. Currently, I will use the release version `20.20200210`

When building custom gadget, I recommend to use stage-snap of pc snap, then customize it further as needed. Please do not use the pc gadget's snapcraft.yaml itself.

xnox:
@anonymouse64 @chrisccoulson @mvo5 @pedronis For me to merge this, I will then have to also adjust bootchains in snapd's bootloader/grub.go right?

Cause now the chain can be either EFI/boot prefix or EFI/ubuntu prefix, in the grubRecoveryModeTrustedAssets..... depending on which revision of pc snap was used.

I guess I should fix the buildability of this snap, independent of the fallback change.

Makefile
```diff
@@ -64,9 +64,22 @@ all:
 	# BIOS boot partition must be defined with an absolute offset.  The
 	# particular value here is 2049, or 0x01 0x08 0x00 0x00 in little-endian.
 	/bin/echo -n -e '\x01\x08\x00\x00' | dd of=pc-core.img seek=500 bs=1 conv=notrunc
-	cp $(SNAPCRAFT_STAGE)/usr/lib/shim/shimx64.efi.dualsigned shim.efi.signed
-	cp $(SNAPCRAFT_STAGE)/usr/lib/grub/x86_64-efi-signed/grubx64.efi.signed grubx64.efi
+	# We must pull in dualsigned shim & grub with UC20 signature
+	# Do it by hand, as snapcraft doesn't have support for PPA archives yet
+	# And yet people try to rebuild this gadget snap
```
anonymouse64, line 69: but we do now have ppa support in snapcraft.yaml proper, it is non-experimental and in the stable channel available everywhere

anonymouse64 (changes requested):
We can use external repositories (i.e. PPA's) in snapcraft.yaml properly now, I think we should do that instead

Makefile
anonymouse64: example: https://github.com/snapcore/pi-gadget/pull/73

xnox:
@mvo5 @chrisccoulson @sil2100 i thought we can use shim fallback by default on first boot. Will that still work correctly with resealing? This pull request has been pending for UC20 since forever. Should we try to land this into UC22?

Makefile
xnox: Actually we can't use package-repositories here, because we must pin to UC20 signing key binaries, and the version numbers of those in the UC20 ppa and the primary archive can be the same. Or like primar archive be ahead. And there is no pinning support. Thus direct pulls like that are best at the moment.

xnox:
I feel this should be added to UC22 first and see how much stuff breaks.

Makefile
anonymouse64: does package-repositories not support pinning to the version you need and specifying to use the version from the PPA over the main archive? if not, sounds like it should support that

xnox: it does not currently have an option to specify a pin priority.

xnox:
Requires https://github.com/snapcore/snapd/pull/11448

xnox:
i'm not working on this currently.

PR #53, opened 2021-06-16
Adding additional kernel command line parameters

bugraaydogar:
Snapd Team introduced a way to customize the kernel command line parameters. This is a reference implementation and I believe it is beneficial to keep it in the upstream.

https://forum.snapcraft.io/t/customising-uc20-kernel-command-line-arguments/24370

### DIFF
--- a/cmdline.extra
+++ b/cmdline.extra
@@ -1,0 +1,1 @@
+custom.option=1
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -23,6 +23,9 @@ parts:
     source: .
     build-packages:
       - grub-common
+    override-build: |
+      snapcraftctl build
+      cp cmdline.extra $SNAPCRAFT_PART_INSTALL/
     plugin: make
     after: [grub-prepare]
 


xnox:
> Snapd Team introduced a way to customize the kernel command line parameters. This is a reference implementation and I believe it is beneficial to keep it in the upstream.
> 
> https://forum.snapcraft.io/t/customising-uc20-kernel-command-line-arguments/24370

This is a default gadget as used for reference images.... doing the above, will kind of break upgrades and is not useful at all.

Also snapcraft syntax that does override build is not quite nice. One can simply stage-files directly.

Can you think of a more useful thing to customize that we could ship by default? Or should our stock cmdline.extra be commented out such that it has no effect?

For most part, when people customize gadget I expect them to stage-snap pc as a whole, add new files. I wonder if we should build & publish such a customized-pc gadget showing the best practice on the available customizations.

bugraaydogar:
Hmm, I have seen that people directly fork this repo and customize it directly according to their needs without staging pc snap. This was the reason that I wanted to make the changes in this repo. Probably, the following repo is a better place for the changes. We could even use the following repo to show best practice on the available customization.

https://git.launchpad.net/~lyoncore-team/lyoncore-snaps/+git/pc-amd64-gadget/tree/snapcraft.yaml?h=20

ogra1:
Rebuilding the gadget for your devices is a hard requirement for all brand store users, all customer facing documentation since UC16 suggests forking the pc or the pi gadgets from here to add your serial-vault/store registration bits ...

but i think describing how to use ```cmdline.extra``` in the README should be sufficient vs introducing a non-used cmdline option by default

xnox:
> Rebuilding the gadget for your devices is a hard requirement for all brand store users, all customer facing documentation since UC16 suggests forking the pc or the pi gadgets from here to add your serial-vault/store registration bits ...
> 

Can they still stage, same name gadget, from the global store?

As in "Building PC, stage gadget PC, add custom files, upload to brand store"

anonymouse64:
> Can they still stage, same name gadget, from the global store?

They can, but it's tricky to get notifications that you need to re-build, we don't (yet) have the same "out of date stage-packages" system for "out of date stage-snaps" notifications built. Regardless of this, it would still be good to document how to use cmdline.extra or cmdline.full here

bugraaydogar:
Okay, so I will remove this pull request and create a new one that modifies the README to explain how to modify the kernel command line parameters. Thanks for the discussions.

clorichel:
For those ending up here from search engines 🤓

On the `22` branch, a `Makefile` is used, so you do NOT have to override the build as suggested by this PR. Assuming you want to add a `cmdline.extra` file containing:

```
quiet splash
```

or really anything else that suits you. Placing this file on the top level of this repo is not enough, you also want to add `install -m 644 cmdline.extra $(DESTDIR)/` on the `install` section of the `Makefile` for the built snap to have that file on its top level too, which is where it's actually picked up by Ubuntu Core.

PR #54, opened 2021-10-27
Classic experiments

sil2100:
...this would still require changes in livecd-rootfs, probably. Since I don't think we have grub.cfg there? Still thinking about it.


sil2100:
This needs the rootfs to include a `grub.cfg`, meaning requires https://code.launchpad.net/~ubuntu-core-dev/livecd-rootfs/+git/livecd-rootfs/+merge/410949 to land as well.

murraybd:
This looks good to me, thanks for sorting it out.

PR #55, opened 2022-02-22
22 focal with fallback

xnox:
This branch adds reproducible building;
using *focal* binaries;
adds support for fallback.

TODO: test that encrypted boot works


xnox:
This breaks encrypted installs with unable to find recovery mode trusted assets https://bugs.launchpad.net/snapd/+bug/1962182

xnox:
Requires https://github.com/snapcore/snapd/pull/11448

PR #56, opened 2022-02-22
22 jammy with fallback

xnox:
This branch builds on top of https://github.com/snapcore/pc-amd64-gadget/pull/55

This branch upgrades to jammy shim & grub binaries.

TODO: test that encrypted boot works

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -29,9 +29,17 @@ volumes:
           edition: 2
         content:
           - source: grubx64.efi
-            target: EFI/boot/grubx64.efi
+            target: EFI/ubuntu/grubx64.efi
           - source: shim.efi.signed
-            target: EFI/boot/bootx64.efi
+            target: EFI/BOOT/BOOTX64.efi
+          - source: fbx64.efi
+            target: EFI/BOOT/fbx64.efi
+          - source: mmx64.efi
+            target: EFI/BOOT/mmx64.efi
+          - source: BOOTX64.CSV
+            target: EFI/ubuntu/BOOTX64.CSV
+          - source: shim.efi.signed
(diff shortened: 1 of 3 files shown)


xnox:
Note that #55 breaks things

gadget.yaml
```diff
@@ -29,9 +29,17 @@ volumes:
...
           - source: grubx64.efi
-            target: EFI/boot/grubx64.efi
+            target: EFI/ubuntu/grubx64.efi
           - source: shim.efi.signed
-            target: EFI/boot/bootx64.efi
+            target: EFI/BOOT/BOOTX64.efi
+          - source: fbx64.efi
+            target: EFI/BOOT/fbx64.efi
+          - source: mmx64.efi
+            target: EFI/BOOT/mmx64.efi
```
alfonsosanchezbeato, line 38: Do we really want MokManager in the gadget? @anonymouse64 @pedronis @mvo5

anonymouse64:
@xnox should this be targeting the 20 branch or the 22 branch? It currently seems to be targeting the 20 branch which seems somewhat unexpected?

xnox:
> @xnox should this be targeting the 20 branch or the 22 branch? It currently seems to be targeting the 20 branch which seems somewhat unexpected?

well spotted! fixed.

And needs https://github.com/snapcore/snapd/pull/11448

gadget.yaml
xnox: I think there were requests for `mokutil` in core before, and also requests to thus ship mokmanager in the gadget. But I cannot find that any more. Maybe ping Field? CE? @vorlonofportland do you recall any requests or need for mokmanager on Core?

PR #59, opened 2022-04-04
Revert focal grub changes for the new UC-signed grub2 

sil2100:
Now that we have a new grub2 signed, I'm reverting xnox's earlier change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,25 +56,16 @@ GRUB_MODULES = \
 	video
 
 all:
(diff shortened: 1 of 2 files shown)


xnox (changes requested):
Nope, that's not the right thing to do......

we should not be reverting packaging improvements, that guarantee pulling correct shim from the correct ppa.

the required change is to replace focal with jammy basically, and drop pulling shim.....

PR #60, opened 2022-04-05
Upgrade pulling shim-signed & grub2-signed from jammy

xnox:
Drop pulling shim, as it is now an empty package.


sil2100 (approved):
Works indeed, looking good.

PR #61, opened 2022-05-17
Unbreak core22 gadget

xnox:
gadget.yaml is no longer automatically staged.


mvo5 (approved):

xnox:
travis.org is gone; we should switch to travis.com or something else....

ogra1 (commented):
could that be done from proper snapcraft.yaml syntax instead of shelling out to make once again ? 
our gadgets are tried to be used as reference for customers and it is a lot of work to always having to help them to translate these ancient hacks into proper snapcraft code ...

PR #62, opened 2022-06-09
snap: increase size of ubuntu-save to 32Mb

mvo5:
Currently the ubuntu-save partition needs to be bigger than 32Mb 
for devices that have 4k sector sizes. This is because of a bug in
how snapd calls mkfs [1] and because of a chance in cryptsetup [2].

With the bugfix for [1] the usable space of the ubuntu-save is very
small (just 3.8Mb) and much less than the 6.9Mb that UC20 has for
512b sector sizes (there is no support for 4kb sectors in the uc20 
cryptsetup). So to ensure there is enough usable space on ubuntu-save
this commit increases the size of ubuntu-save to 32Mb which translates
to around 18Mb of usable space.

With 33 we have:
```
mvo@ubuntu:~$ df -h /run/mnt/ubuntu-save/
Filesystem                                                    Size  Used Avail Use% Mounted on
/dev/mapper/ubuntu-save-4f616498-7704-487b-939c-90986debe7f7   21M  120K   19M   1% /run/mnt/ubuntu-save
```

For reference on UC20 we have (with 16mb):
```
mvo@ubuntu:~$ df -h /run/mnt/ubuntu-save/
Filesystem                                                    Size  Used Avail Use% Mounted on
/dev/mapper/ubuntu-save-498a65bb-30ce-492c-a8aa-301adb79228e  7.8M  187K  6.9M   3% /run/mnt/ubuntu-save
```

[1]  https://bugs.launchpad.net/snappy/+bug/1968356
[2] https://gitlab.com/cryptsetup/cryptsetup/-/merge_requests/135

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -49,7 +49,7 @@ volumes:
         role: system-save
         filesystem: ext4
         type: 83,0FC63DAF-8483-4772-8E79-3D69D8477DE4
-        size: 16M
+        size: 64M
       - name: ubuntu-data
         role: system-data
         filesystem: ext4
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -49,7 +49,7 @@ volumes:
         role: system-save
         filesystem: ext4
         type: 83,0FC63DAF-8483-4772-8E79-3D69D8477DE4
-        size: 64M
+        size: 32M
       - name: ubuntu-data
         role: system-data
         filesystem: ext4


sil2100 (approved):
I have no particular preferences here, 64M seems fine. I'd like to hear from others though.

xnox:
http://manpages.ubuntu.com/manpages/jammy/man8/cryptsetup.8.html#:~:text=%2D%2Dsector%2Dsize%20%3Cbytes%3E is quite an informative piece of text, which implies forcing LUKS2 to use --sector-size 4096 everywhere is dangerous.

xnox:
If we don't care about the most optimal performance of ubuntu-save partition given its small size, would it be prudent to change snap-bootstrap to create ubuntu-save LUKS2 volume with sector-size 512? That way existing size of the partition will be suitable, and will result in a reasonably sized ext4 fs as well.

Note this is just I guess, I have not tested if it actually allowed to force cryptsetup --sector-size 512 on a 4k drive followed by 1k ext4 for ubuntu-save.

mvo5:
After some discussion I think the best option is:
1. switch this to 32Mb
2. merge https://github.com/snapcore/snapd/pull/11857

Given that 4k block size is not super common yet I think we can release with (1) only but need to update the image soon once (2) is in.

PR #63, opened 2022-07-08
snapcraft.yaml: allow building outside of launchpad with old binaries

### LINKED ISSUE #57: Secure boot installation of uc22 does not work since 22-0.2
I could still install on qemu with secure boot and tpm2, UC22 with pc gadget 22-0.1 (118).

Since 22-0.2 (119), that does not work anymore. After reboot, snap-bootstrap cannot unlock the encrypted partitions.

Also grub prints `/EndEntire` two times. Though I do not think it is related, but we do not need this messages in the console.

alfonsosanchezbeato:
The measurements in 2.06 grub are currently incorrect, force using
focal binaries, without requirement to specify archive in launchpad
/+snap/ page such that one can build the gadget locally correctly too.

Also remove all unused dependencies. And it is no longer required to
stage build-deps, instead of build-depend on them, as they are no
longer conflicting with anything installed on the host.

Fixes #57

Backport from 00ea33892ff0b258d2e4136d7b911a9c8ce0605d

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,16 +56,25 @@ GRUB_MODULES = \
 	video
 
 all:
-	dd if=$(SNAPCRAFT_STAGE)/usr/lib/grub/i386-pc/boot.img of=pc-boot.img bs=440 count=1
+	dd if=/usr/lib/grub/i386-pc/boot.img of=pc-boot.img bs=440 count=1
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato:
Part of this was also in https://github.com/snapcore/pc-amd64-gadget/pull/52

Meulengracht (approved):
Looks good, thanks!

xnox (approved):

PR #64, opened 2022-07-11
Support kernel with EFI runtime turned off

xnox:
Kernel configuration supports turning off EFI runtime services on boot. Meaning, the kernel is booted with UEFI Secureboot and yet it doesn't allow access to efivarfs or any EFI services.

One such common kernel configuration is for Ubuntu Realtime Kernel. That is driven by the fact that EFI runtime services are unpredictable in their timings (reading and writing efivars can trigger a blocking EFI garbage collection of unknown amount of time).

On the other hand, Ubuntu Core sealed full disk encryption is a key feature.

Specify `efi=runtime` kernel command line option, such that stock EFI gadget can be used with realtime-kernel and have full disk encryption out of the box.

Note that snap refresh of gadget & kernel snaps, may result in realtime latency spikes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,5 +77,6 @@ all:
 install:
 	install -m 644 pc-boot.img pc-core.img shim.efi.signed grubx64.efi $(DESTDIR)/
 	install -m 644 grub.conf $(DESTDIR)/
+	install -m 644 cmdline.extra $(DESTDIR)/
 	install -d $(DESTDIR)/meta
 	install -m 644 gadget.yaml $(DESTDIR)/meta/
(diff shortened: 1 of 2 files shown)


xnox:
Alternative is to upgrade static-cmdline snippets in https://github.com/snapcore/snapd/blob/master/bootloader/assets/grub.go#L26

xnox:
Third option to report it as ubuntu-realtime bug and say that efi runtime should be turned on in realtime kernel, because modern computing needs it.

xnox:
this shell be fixed in the RT kernel itself.

PR #68, opened 2022-08-23
Increase boot partition size for classic gadget

jawn-smith:
The following math was done to come up with the number of 512:

```
In [1]: fudge_factor = 2

In [2]: num_copies = 1 # we should only need one copy of the kernel for pc-amd64 images

In [3]: kernel_size = 11

In [4]: initrd_size = 169

In [5]: other_boot_assets = 22 # /boot/grub and /boot/EFI

In [6]: boot_size = fudge_factor * (num_copies * ( kernel_size + initrd_size +other_boot_assets))

In [7]: boot_size
Out[7]: 404
```

Then rounding up from 404 to a nice power of 2 we get 512.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -18,7 +18,7 @@ volumes:
         type: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
         filesystem: vfat
         filesystem-label: system-boot
-        size: 50M
+        size: 200M
         content:
           - source: grubx64.efi
             target: EFI/boot/grubx64.efi
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -18,7 +18,7 @@ volumes:
         type: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
         filesystem: vfat
         filesystem-label: system-boot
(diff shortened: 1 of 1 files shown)


xnox:
```
num_copies = 1 # we should only need one copy of the kernel for pc-amd64 images
```

I don't think is true, since snapd does unpack two kernel.efi's during upgrade to ubuntu-boot partition for a/b booting. Or has this been changed somewhere else for this mode?

jawn-smith:
@xnox thanks for taking a look! Is a/b booting something that is relevant for classic images? This PR is for the classic branch, which has a `system-boot` partition instead of `ubuntu-boot`.

That being said, I've looked into how current amd64 preinstalled server images are set up. It seems only /boot/efi is mounted on a separate partition, which doesn't include the kernel and initrd. The kernel and initrd come from the `cloudimg-rootfs` partition. This is not how the `ubuntu-image` redesign was planned to work, so I need to tag up with @sil2100 to see if the expectations of `ubuntu-image` always placing a kernel and initrd in the `system-boot` partition need to be reconsidered.

PR #66, opened 2022-08-01
Build from snapcraft.yaml and use ucdev PPA

alfonsosanchezbeato:
Remove makefile and do special steps from an override-build section
instead.

Take now signed grub and shim form the ucdev/uc-staging-ppa PPA, using
package-repositories. Now we use epochs in the version of
grub-efi-amd64-signed and shim-signed packages, so the build should
always take the ones from the PPA instead of those in the archive, as
`2:<version>` will always be greater than the package version in the
latter. Nonetheless, do an additional static check for this.

Forward-ported from https://github.com/snapcore/pc-amd64-gadget/pull/65

PR #69, opened 2022-10-31
prepare-device: implement reference prepare-device hook

alexclewontin:
Snapcraft.io points to this as the "reference" implementation of the amd64 gadget. However, no prepare-device hook makes this an incomplete reference, especially for private store operators

The proposed prepare-device hook would not change the behavior of the default PC gadget. However, it could then be forked and trivially modified to use for private store authentication.

### DIFF
--- a/snap/hooks/prepare-device
+++ b/snap/hooks/prepare-device
@@ -1,17 +1,23 @@
 #!/bin/sh
 
-set -eux
+# This file is included for reference purposes. If MODEL_APIKEY is not set
+# in snapcraft.yaml (as is the default), it is a no-op. For more about how
+# to use this script to connect a device to an IoT App Store, see
+#  - https://ubuntu.com/core/services/guide/serial-vault-overview
+#  - https://ubuntu.com/core/services/guide/connecting-devices
+#  - https://snapcraft.io/docs/gadget-snap#heading--prepare
+
+set -eu
+
+if [ -z "$MODEL_APIKEY" ] ; then
+    exit 0
+fi
 
 exec >> $SNAP_COMMON/prepare-device-hook.log 2>&1
 
-# TODO: define your model's API key here
-# See https://ubuntu.com/core/services/guide/serial-vault-overview
-# for instructions on how to generate an API key
-# DO NOT check this API key into a publicly accessible VCS
-MODEL_APIKEY=""
+# If you are forking and building your own gadget:
+# implement your preferred way of generating a serial number for this device here
+snapctl set registration.proposed-serial="\"$(date -u)\""
 
-if [ -n "$MODEL_APIKEY" ] ; then
-    snapctl set registration.proposed-serial="\"$($SNAP/usr/bin/uuidgen)\""
-    snapctl set device-service.url="https://serial-vault-partners.canonical.com/v1/"
-    snapctl set device-service.headers="{\"api-key\": \"$MODEL_APIKEY\"}"
-fi
+snapctl set device-service.url="https://serial-vault-partners.canonical.com/v1/"
+snapctl set device-service.headers="{\"api-key\": \"$MODEL_APIKEY\"}"
--- a/snap/hooks/prepare-device
+++ b/snap/hooks/prepare-device
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -eux
+
+exec >> $SNAP_COMMON/prepare-device-hook.log 2>&1
+
+# TODO: define your model's API key here
+# See https://ubuntu.com/core/services/guide/serial-vault-overview
+# for instructions on how to generate an API key
+# DO NOT check this API key into a publicly accessible VCS
+MODEL_APIKEY=""
+
+if [ -n "$MODEL_APIKEY" ] ; then
+    snapctl set registration.proposed-serial="\"$($SNAP/usr/bin/uuidgen)\""
+    snapctl set device-service.url="https://serial-vault-partners.canonical.com/v1/"
+    snapctl set device-service.headers="{\"api-key\": \"$MODEL_APIKEY\"}"
+fi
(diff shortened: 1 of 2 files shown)


xnox:
however store operators should not fork pc-amd64-gadget, as that's hard to maintain. Instead they should really stage-snap: pc-gadget instead.

It probably is best to create a new branch here that does stage-snap: pc-gadget and has reference hooks to build a store specific pc-gadget.

alexclewontin:
> however store operators should not fork pc-amd64-gadget, as that's hard to maintain. Instead they should really stage-snap: pc-gadget instead.
> 
> It probably is best to create a new branch here that does stage-snap: pc-gadget and has reference hooks to build a store specific pc-gadget.

@xnox problem with that is that all the parts of the gadget are fairly coupled. Store operators still need their own gadget.yaml file (to set default snap config, some autoconnections), and then gadget.yaml directly references a bunch of the other pieces of the snap (the various boot assets and the like). So using the generic gadget as a stage snap and manually making sure your gadget.yaml picks up all the necessary changes is probably _less_ maintainable than simply rebasing your source changes on top of our source changes.

Also stage snaps don't get USNs, but debs do. So when #66 gets merged, you'll actually get more security notifications if you publish your own forked version than if you stage the snap

snap/hooks/prepare-device
```diff
@@ -0,0 +1,17 @@
...
+
+set -eux
+
+exec >> $SNAP_COMMON/prepare-device-hook.log 2>&1
+
+# TODO: define your model's API key here
+# See https://ubuntu.com/core/services/guide/serial-vault-overview
+# for instructions on how to generate an API key
+# DO NOT check this API key into a publicly accessible VCS
+MODEL_APIKEY=""
```
dilyn-corner: Considering this line simply replaces `MODEL_APIKEY` with an empty string, we should probably use something like `${MODEL_APIKEY:-""}`

alexclewontin: This runs as a hook, so where else would MODEL_APIKEY get set?

alexclewontin: Oh I guess you could set as an env var in snapcraft.yaml

alexclewontin: Fair enough

snapcraft.yaml
```diff
@@ -17,4 +17,16 @@ parts:
       - grub-pc-bin
       - grub-common
     plugin: make
+  uuidgen:
+    plugin: nil
+    stage-packages: [ uuid-runtime ]
```
alfonsosanchezbeato: I would prefer to avoid staging any additional packages not strictly needed by the gadget - maybe something like creating the serial by looking at `/dev/urandom` or by running `date` as we had in the past.

snap/hooks/prepare-device
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -eux
+
+exec >> $SNAP_COMMON/prepare-device-hook.log 2>&1
```
alfonsosanchezbeato: I'd prefer avoiding any file creation unless this runs for the real use case, and even to return early. Something like
```
if [ -z "$MODEL_APIKEY" ]; then
    exit 0
fi

exec >> $SNAP_COMMON/prepare-device-hook.log 2>&1
...
```

snap/hooks/prepare-device
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -eux
```
alfonsosanchezbeato: `-x` probably can be removed

snapcraft.yaml
```diff
@@ -17,4 +17,16 @@ parts:
...
+    stage: [ usr/bin/uuidgen, usr/share/doc/uuid-runtime/copyright ]
 
+hooks:
+  prepare-device:
+    environment:
+    # TODO: define your model's API key here
```
alfonsosanchezbeato: We should not have a `TODO` as it applies only if you are forking the gadget. Something like "If you are forking the gadget you can define...".

snapcraft.yaml
```diff
@@ -17,4 +17,16 @@ parts:
...
+    # See https://ubuntu.com/core/services/guide/serial-vault-overview
+    # for instructions on how to generate an API key
+    # DO NOT check this API key into a publicly accessible VCS
```
alfonsosanchezbeato: Nitpick^2: I would prefer this comment indented to the right a couple of spaces.

snap/hooks/prepare-device
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
alfonsosanchezbeato, line 1: A header saying that this file is included just for reference purposes and some link would be great.

alfonsosanchezbeato (commented):
I do not oppose merging this, I think that it can help people that need to fork pc-gadget. I have some comments though.

snapcraft.yaml
alexclewontin: Switched to `date -u`

snapcraft.yaml
alexclewontin: I think this is what you were asking, but let me know if it should go further

snap/hooks/prepare-device
alexclewontin: Added, let me know if that covers it/if it should be more or less/any nitpicks on wording

alfonsosanchezbeato (approved):
LGTM, thanks for the changes

dilyn-corner:
LGTM :)

dilyn-corner (approved):

dilyn-corner:
Is there an update on merging this? @alfonsosanchezbeato

alfonsosanchezbeato:
Thanks, merged now

PR #70, opened 2022-11-16
Copy shim in EFI/ubuntu/shimx64.efi

valentindavid:
fwupd expects a copy of shim to be available there.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -32,6 +32,8 @@ volumes:
             target: EFI/boot/grubx64.efi
           - source: shim.efi.signed
             target: EFI/boot/bootx64.efi
+          - source: shim.efi.signed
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato (approved):
LGTM, as we have the same in classic systems too

xnox:
@pedronis had some restrictions on this before. I can't remember if all targets must be a unique set, or if all sources must be a unique set, or both. I would feel safer with a second source copy of shim under a different name inside the gadget.

xnox (approved):
This does not change the bootchain as used for booting, and there are still uniquely named targets. Thus issues raised in https://github.com/snapcore/snapd/pull/11437#issuecomment-1073860312 do not apply as the names are still unique. This thus unblocks platform features that others need.

PR #71, opened 2022-11-16
update gadget.yaml to reflect location and size of existing partitions

mwhudson:
Running sfdisk on the target system gives output like this:

image002

This is my attempt to make gadget.yaml match this (slightly surprising) output.


xnox:
If we are going to drop bios boot mbr, we shall also exclude it from the gadget.

somethinng like `stage: - -pc-boot.img`

xnox:
i will look into it.

PR #74, opened 2022-11-24
Remove initial offset for the first partition

alfonsosanchezbeato:
This is causing problems in qemu but also on real HW due probably to
LP: #1929087.


mvo5 (approved):
Thanks!

mwhudson:
Oh hm we should add an explicit offset to the recovery partition, not sure why I didn't do that in #71

mwhudson:
FWIW I have curtin fixes that mean we could drop this change now, if desired.

PR #75, opened 2022-11-24
Revert role to system-seed

alfonsosanchezbeato:
This is needed so we mount this partition in /run/mnt/ubuntu-seed. In the future the behaviour of system-seed-null should be the same, so this is a workaround. Mounting is needed as we check binaries in ubuntu-seed to find out if sealing the disk key is necessary.


pedronis (approved):
this is ok but means we will potentially have to write extra code to deal with gadget updates that change the role again

PR #76, opened 2022-11-24
add an explicit offset to the recovery partition

mwhudson:
I haven't tested this (not really sure how to in a sensible amount of time), but I don't think the current version will put the recovery partition in the right place


mvo5 (approved):

mvo5:
Fwiw, I tested this with spread and it seems to work, thank you!

PR #77, opened 2022-11-24
gadget.yaml: rename EFI System partition to ubuntu-seed

mvo5:
This will make the install work but it's not quite correct, this is not really a seed partition :-/ But it unblocks the install

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -4,7 +4,7 @@ volumes:
     bootloader: grub
     schema: gpt
     structure:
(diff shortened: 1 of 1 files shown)


pedronis:
without understanding which bit requires this I'm not very keen on this

mvo5:
With a slightly fresher mind I see now that the error  https://paste.pics/K0P29 comes from muinstaller so hopefully we don't need this

PR #78, opened 2023-01-24
Set system-seed-null for ESP partition  and enable splash

alfonsosanchezbeato:


gadget.yaml
```diff
@@ -5,7 +5,7 @@ volumes:
     schema: gpt
     structure:
       - name: EFI System partition
-        role: system-seed
+        role: system-seed-null
```
xnox, line 8: which snapd / snapcraft support that new role name? is it safe to merge now, or will it require edge channel to build the gadget snap and the image?

Note for image building I believe we need this role supported in the snapd.deb snap-prepare etc.

xnox: checked snapd, core-initrd, pc-kernel snaps. snapd snap and rmadison (in proposed)

PR #80, opened 2023-02-22
Make the gadget compatible with arm64

alfonsosanchezbeato:


gadget.yaml
```diff
@@ -1,24 +1,9 @@
...
-          - image: pc-boot.img
-      - name: BIOS Boot
-        type: DA,21686148-6449-6E6F-744E-656564454649
-        size: 1M
-        offset: 1M
-        offset-write: mbr+92
-        update:
-          edition: 2
-        content:
-          - image: pc-core.img
```
valentindavid, line 21: Nice. That did not work anyway.

valentindavid, line 21: I do not think that works on x86_64. Can we just remove it for all?

xnox: It does work..... it boots and prints a message saying "please boot in EFI mode" =)

xnox (approved):

gadget.yaml
alfonsosanchezbeato: Why do you think it does not work? It is expected to be called for DOS style disks afaiu.

xnox: Actually in the current grub-recovery.conf i don't see a test for efi platform boot, which used to print a message "please boot in EFI mode" https://github.com/snapcore/snapd/blob/2b72f0457b8b8d35906784a4bfeceee1cf49bbcd/bootloader/assets/data/grub-recovery.cfg#L33

Like we used to have somewhere before.... hmmm.

valentindavid: Do we need all this just to print a message? I would think we could print that message is in less than 440 bytes. Maybe we can replace by a really simpler MBR.

alfonsosanchezbeato: In any case, that would be for another PR to branch `22`

xnox: Indeed BIOS boot is not available on arm64 so all of these things should be discussed elsewhere.

valentindavid (approved):

PR #81, opened 2023-02-22
Remove ubuntu-boot and replace with a simpler mbr

valentindavid:


Makefile
```diff
@@ -55,15 +55,7 @@ GRUB_MODULES = \
 	true \
 	video
 
-all:
```
xnox, line 58: everything above can be removed as well given this change.

xnox (approved):
In the Makefile, lists of grub modules to vendor no longer needed and should be removed. Only static grub.efi binary is needed now.

In snapcraft.yaml you can then also drop "build-packages"   [ grub-pc-bin, grub-common ]

xnox (changes requested):
See above requests to remove even more things given this change.

Makefile
valentindavid: Done.

legacy-bios/mbr.s
```diff
@@ -0,0 +1,28 @@
+        .code16
+        .section .text
+        .globl start
+
+start:
```
Meulengracht, line 5: Usually it's good practice to clear the code-segment as the first thing in bootloaders, you can't trust that the code-segment is valid. Adding the following lines is enough (sorry for my intel syntax):

```
start:
	jmp 0x0:entry

entry:
```

legacy-bios/mbr.s
```diff
@@ -0,0 +1,28 @@
...
+        xorw %ax, %ax
+        movw %ax, %ss
+        movw %ax, %ds
+        movw $0x7c00, %sp
```
Meulengracht, line 12: While setting up segments it's a good idea to do this with interrupts disabled (don't want to modify SS without SP and then get an interrupt in-between)

```suggestion
        cli
        xorw %ax, %ax
        movw %ax, %ss
        movw %ax, %ds
        movw $0x7c00, %sp
        sti
```

legacy-bios/mbr.s
valentindavid: Thank you. That was `jmp $0, $entry`.

legacy-bios/mbr.s
valentindavid: Done.

legacy-bios/mbr.s
```diff
@@ -0,0 +1,31 @@
...
+
+start:
+        jmp $0, $.entry
+.entry:
+        cli
+        xorw %ax, %ax
+        movw %ax, %ss
+        movw %ax, %ds
+        movw $0x7c00, %sp
+        sti
```
Meulengracht, line 13: Last comment, and then it's a +1 from me, I missed this on initial review, but add a 'cld' instruction before printing the string. The 'cld' instruction clears the direction flag which is a good thing to do initially as we don't know the state of the direction flag when loaded. This affects how  `lodsb` loads string data.

```suggestion
        sti
        cld
```

mvo5 (approved):
Nice, thank you!

xnox (approved):

legacy-bios/mbr.s
valentindavid: Done!

Meulengracht (approved):
Thanks for the changes, LGTM!

alfonsosanchezbeato (approved):
Very nice, LGTM

PR #83, opened 2023-03-01
Re-add an empty bios boot partition to allow updates

valentindavid:


xnox (approved):
sadness and despair.

xnox (commented):
probably was my mistake when i dind't bump mbr partition edition to 2 as well, when bumping bios & ubuntu-seed partitions. And sort of moot and pointless anyway now.

alfonsosanchezbeato (approved):
LGTM

PR #84, opened 2023-03-15
Update partition layouts for 22-xinyi-pc-classic

medicalwei:
Update the partition sizes according to the info from our customer.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -8,7 +8,7 @@ volumes:
         role: system-seed-null
         filesystem: vfat
         type: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
-        size: 248951808
+        size: 786432000
         offset: 1048576
         update:
           edition: 2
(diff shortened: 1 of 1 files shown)


gadget.yaml
```diff
@@ -19,16 +19,21 @@ volumes:
...
       - # Recovery partition
         type: E3C9E316-0B5C-4DB8-817D-F92DF00215AE
-        size: 8192000000
-        offset: 250000384
+        size: 12884901888
+      - name: ubuntu-seed
+        role: system-seed
+        filesystem: ext4
+        type: 0FC63DAF-8483-4772-8E79-3D69D8477DE4
+        size: 1200M
```
xnox: having partition ubuntu-seed as ext4 is unusual. And having it as not ESP is unusual. I am not sure if it is valid to have ubuntu-seed with a role system-seed, whilst ESP is marked as system-seed-null.

Has this been checked with snapd team? @alfonsosanchezbeato

xnox (commented):
I would like to request ubuntu-core/snapd team to review this change.

xnox:
@alfonsosanchezbeato @pedronis @mvo5 do things proposed here adhere to gadget specifications?

mwhudson:
I think it would be best to get the new sizes landed asap and then have a proper conversation about ubuntu-seed in a separate PR.

xnox:
> I think it would be best to get the new sizes landed asap and then have a proper conversation about ubuntu-seed in a separate PR.

can i please have a new PR, which just updates the sizes of existing things and/or gaps&offsets without introducing `role: system-seed` (the gap is there currently, right?)

medicalwei:
I updated the PR so that only offsets and sizes are changed.  The gap is now 1200MiB between recovery and ubuntu-boot

xnox (approved):

PR #85, opened 2023-03-15
Fix compatibility with old and new shim naming

jawn-smith:
Oops.


sil2100 (approved):
Ok, this looks fine, I like the very generic way of handling this. Remember that for much more complex differences series-by-series we currently always pass `SERIES` when running make on the gadget. But this way here is very convenient already.

PR #89, opened 2023-04-24
Pull shim and grub from the Ubuntu archive

alfonsosanchezbeato:
We do not want the UC certificate in the xinyi gadget, so we pull shim/grub from archive instead of staging from the pc gadget. This also cleans up files that were pulled and not really needed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: xinyi-pc-classic
-version: '22-0.4'
+version: '22.04-0.1'
 type: gadget
 base: core22
 summary: Xinyi PC Classic gadget
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -8,6 +8,10 @@ description: |
 confinement: strict
 grade: stable
 
+# Min version to support fully FDE on classic
+assumes:
+  - snapd2.59.3
+
 parts:
   boot:
     plugin: nil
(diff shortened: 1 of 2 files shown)


snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: xinyi-pc-classic
-adopt-info: pc
+version: '22-0.4'
```
xnox: Imho version should be set to something meaningful, like shim-signed + grub-efi-amd64-signed versions, using a version part. Given this is now following the archive, and doesn't have Ubuntu Core meaning.

This is a minor comment, and can be fixed up later.

xnox (approved):
no idea how to ensure ordering with snapd upgrade.

alfonsosanchezbeato:
> no idea how to ensure ordering with snapd upgrade.

We can add:

```
assumes:
  - snapdX.XX
```

once we have a snapd release containing the secboot fix we can add this to the PR. Meanwhile I've converted this to a draft, to avoid accidental releases.

snapcraft.yaml
alfonsosanchezbeato: I think having a version as what you say would be too long, so I have changed it to something using 22.04 so it is clearer this is for classic. But it is a good idea that we know where things came from, so I've now primed `usr/share/doc/` for shim/grub.

xnox (approved):

alfonsosanchezbeato:
Assumes added now

PR #90, opened 2023-05-11
Copy grub to EFI/ubuntu/ folder in seed partition

alfonsosanchezbeato:
Now that we have EFI/ubuntu/shimx64.efi, some BIOS seem to try to boot with it as first option, and then the boot fails as grub is not found. Copy it to EFI/ubuntu/ to fix that.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -31,6 +31,8 @@ volumes:
         content:
           - source: grubx64.efi
             target: EFI/boot/grubx64.efi
+          - source: grubx64.efi
+            target: EFI/ubuntu/grubx64.efi
           - source: shim.efi.signed
             target: EFI/boot/bootx64.efi
           - source: shim.efi.signed


alfonsosanchezbeato:
Waiting for confirmation from fw vendors on how this works exactly.

vorlonofportland:
Yes, this, uh, is a violation of the UEFI spec if they're booting that path in preference over the removable device path, without it having been configured via firmware variables

alfonsosanchezbeato:
It has been confirmed that this is the behavior for Lenovo and Dell. But closing this one as for the moment the better solution is to revert copying shim to `EFI/ubuntu`, see https://github.com/snapcore/pc-gadget/pull/91.

LaiderLai:
After a meeting with the Lenovo team.
Confirmed the Lenovo BIOS logic as below:
if EFI/<os name>/shimx64.efi exists, then it has higher priority than EFI/boot/bootx64.efi to create UEFI NVRAM variable "BootOrder/[Ubuntu]" to be the boot option target.
If there is no EFI/<os name>/shimx64.efi, then default use EFI/boot/bootx64.efi to create UEFI NVRAM variable "BootOrder/[UEFI OS] to be the boot option target"

The <os name> can be ubuntu, fedora, ..., etc.
The design reason is stock Ubuntu (Ex. Ubuntu desktop) and Fedora, ..., the Linux OS distribution deploys shimx64.efi under the EFI/<os name>/.

And I also found another BIOS have the same logic from a Dell certified device (EGW3200).

PR #91, opened 2023-05-12
Revert "Copy shim in EFI/ubuntu/shimx64.efi"

alfonsosanchezbeato:
It has been found out that some fw vendors (Lenovo and Dell) run EFI/ubuntu/shimx64.efi with higher priority than EFI/BOOT/BOOTX64.EFI, which makes the system unable to boot as there is not EFI/ubuntu/grubx64.efi to jump to from the shim.

The solution is for the moment to revert copying around the shim to EFI/ubuntu, as we cannot simply copy grub to the same place as that could break measurements easily (the device path is part of the EFI_IMAGE_LOAD_EVENT measured in PCR 4) so devices with TPM backed FDE would not boot.

To fix this in a clean way so we can put back EFI/ubuntu/shimx64.efi we will need to start writing EFI boot entries on Ubuntu Core installation.

This reverts commit adadcea8d2107953eb3f422fd341767d67e2acf0.

### DIFF
--- a/gadget.yaml
+++ b/gadget.yaml
@@ -33,8 +33,6 @@ volumes:
             target: EFI/boot/grubx64.efi
           - source: shim.efi.signed
             target: EFI/boot/bootx64.efi
-          - source: shim.efi.signed
-            target: EFI/ubuntu/shimx64.efi
       - name: ubuntu-boot
         role: system-boot
         filesystem: ext4


LaiderLai:
After a meeting with the Lenovo team.
Confirmed the Lenovo BIOS logic as below:
if EFI/"os name"/shimx64.efi exists, then it has higher priority than EFI/boot/bootx64.efi to create UEFI NVRAM variable "BootOrder/[Ubuntu]" to be the boot option target.
If there is no EFI/"os name"/shimx64.efi, then default use EFI/boot/bootx64.efi to create UEFI NVRAM variable "BootOrder/[UEFI OS] to be the boot option target"

The "os name" can be Ubuntu, Fedora, ..., etc.
The design reason is stock Ubuntu (Ex. Ubuntu desktop) and Fedora, ..., the Linux OS distribution deploys shimx64.efi under the EFI/"os name"/.

And I also found another BIOS have the same logic from a Dell certified device (EGW3200).

vorlonofportland:
This looks reasonable to me.

The original rationale for creating \EFI\ubuntu\shimx64.efi was that "fwupd expects it there".  Will there be followup to fix fwupd?

vorlonofportland (approved):

xnox:
> This looks reasonable to me.
> 
> The original rationale for creating \EFI\ubuntu\shimx64.efi was that "fwupd expects it there". Will there be followup to fix fwupd?

i believe fwupd snap is exploring to vendor shims from all distros inside its snap, and deploy and use a matching shim itself as needed.

xnox (approved):

valentindavid (approved):

PR #93, opened 2023-05-17
Fix build for UC20 pc gadget

alfonsosanchezbeato:


xnox (commented):
Does this work with local builds, builds in launchpad, and remote-build?

I thought one is not running as root, and thus cannot modify /etc/apt like that. Unless I am wrong / this has changed. Cause this certainly makes things much easier.

alfonsosanchezbeato:
> Does this work with local builds, builds in launchpad, and remote-build?
> 
> I thought one is not running as root, and thus cannot modify /etc/apt like that. Unless I am wrong / this has changed. Cause this certainly makes things much easier.

This builds in LP (note we do already this for pc gadget for 22) and locally (--use-lxd is fine). snapcraft runs as root in containers/VMs.

xnox (approved):

PR #94, opened 2023-05-18
Build amd64 and arm64 with a unified snapcraft.yaml

alfonsosanchezbeato:


alfonsosanchezbeato:
Creating as a draft as this needs snapcraft with support for pinning repositories (version 7.4.0), which is only available in edge channel.

snapcraft.yaml
```diff
@@ -23,57 +24,69 @@ hooks:
...
+
+      # Remove all the bits we do not need, keeping changelogs and copyrights
+      # (using organize/prime is not possible due to different names per arch - x64/aa64)
+      find "$CRAFT_PART_INSTALL"/ -type f,l \
+                -not -path "$SNAPCRAFT_PART_INSTALL"/shim.efi.signed \
+                -not -path "$SNAPCRAFT_PART_INSTALL"/${grub_bin%.signed} \
+                -not -path "$SNAPCRAFT_PART_INSTALL"/usr/share/doc/grub-efi-$CRAFT_TARGET_ARCH-signed/'*' \
+           -and -not -path "$SNAPCRAFT_PART_INSTALL"/'usr/share/doc/shim-signed/*' \
+           -delete
+      find "$SNAPCRAFT_PART_INSTALL"/ -type d -empty -delete
```
valentindavid, line 78: Unfortunately `find` goes top-down. This is a case I always find annoying. So that does not always remove everything. Maybe using `-exec rmdir --ignore-fail-on-non-empty -p {} ';'` instead of `-delete`?

snapcraft.yaml
```diff
@@ -23,57 +24,69 @@ hooks:
...
+      sbverify --list "$shim_path" | grep -E 'Canonical Ltd. Secure Boot Signing \(Ubuntu Core'
+      sbverify --list "$grub_path" | grep -E 'Canonical Ltd. Secure Boot Signing \(Ubuntu Core'
+
+      # Move shim/grub to the expected path
+      install -m 644 "$shim_path" "$CRAFT_PART_INSTALL"/shim.efi.signed
+      install -m 644 "$grub_path" "$CRAFT_PART_INSTALL"/${grub_bin%.signed}
+
+      # Remove all the bits we do not need, keeping changelogs and copyrights
+      # (using organize/prime is not possible due to different names per arch - x64/aa64)
+      find "$CRAFT_PART_INSTALL"/ -type f,l \
```
valentindavid, line 72: I am not that fan of this `find`. Would not be possible to move the stage packages as build packages instead? And just have `install` or `cp` commands for what we want.

valentindavid (commented):
Just some comment for now.

snapcraft.yaml
alfonsosanchezbeato: If using them as build packages we won't get them in the manifest. Also it implies installation of extra package on the container that would make the build a bit slower.

snapcraft.yaml
alfonsosanchezbeato: Could you please share a case where this could happen? I tried this and worked as expected:
```
$ mkdir -p a/b/c
$ mkdir -p a/a/c
$ find . -type d -empty
./a/a/c
./a/b/c
$ find . -type d -empty -delete
$ find .
.
```

valentindavid: Your are right. I missed something in the doc:
```
       -depth Process each directory's contents before the directory itself.  The -delete action also implies -depth.
```

valentindavid (approved):

PR #96, opened 2023-05-31
Proper series arch handling

lool:
This fixes handling of SERIES, and sets up a sane sources.list. I've also added support for using the -proposed pocket but only if PROPOSED is used. This uses chdist instead of pull-lp-debs as devscripts is in main and not ubuntu-dev-tools.

This is similar to the initial approach I took for classic-tegra/classic-arm64, but much more incremental and simpler to follow. This also uses per-architecture gadget.yamls, the same approach used for core22 snaps.

I've verified the make building logic with:
make ARCH=arm64
make ARCH=amd64
and this passes nicely. However, while I have kept updating snapcraft.yaml, I'm not sure how it's being consumed, so it might need further tweaking.


snapcraft.yaml
```diff
@@ -10,6 +10,11 @@ confinement: strict
 icon: icon.png
 
 parts:
+  gadget-yaml:
+    plugin: nil
+    override-build: |
+      # snapcraft reads gadget.yaml from project dir
+      install -m 644 gadget-$SNAPCRAFT_TARGET_ARCH.yaml $SNAPCRAFT_PROJECT_DIR
```
mwhudson: doesn't this need to install it as "gadget.yaml"?

Makefile
```diff
@@ -62,58 +101,77 @@ GRUB_MODULES = \
...
-	# some packages by ourselves.
-ifndef SNAPCRAFT_PROJECT_NAME
-	$(call stage_package,grub-pc-bin)
-	$(call stage_package,grub-efi-amd64-signed)
-	$(call stage_package,shim-signed)
+	# setup chdist APT environment for SERIES-ARCH and run apt update
+	if [ ! -d  $(STAGEDIR)/tmp/chdist ]; then \
+	    chdist -d $(STAGEDIR)/tmp/chdist -a $(ARCH) create $(SERIES)-$(ARCH); \
+	    echo "deb $(ARCHIVE) $(SERIES) main" >$(STAGEDIR)/tmp/chdist/$(SERIES)-$(ARCH)/etc/apt/sources.list; \
+	    echo "deb $(ARCHIVE) $(SERIES)-updates main" >>$(STAGEDIR)/tmp/chdist/$(SERIES)-$(ARCH)/etc/apt/sources.list; \
```
mwhudson, line 128: Should this include -security?

Makefile
```diff
@@ -62,58 +101,77 @@ GRUB_MODULES = \
...
-
-boot:
-	# Check if we're running under snapcraft. If not, we need to 'stage'
-	# some packages by ourselves.
-ifndef SNAPCRAFT_PROJECT_NAME
-	$(call stage_package,grub-pc-bin)
-	$(call stage_package,grub-efi-amd64-signed)
-	$(call stage_package,shim-signed)
+	# setup chdist APT environment for SERIES-ARCH and run apt update
+	if [ ! -d  $(STAGEDIR)/tmp/chdist ]; then \
```
mwhudson, line 124: You only really need this if SERIES or ARCH does not match the environment you're running in? Or is there some reason to do this then too?

mwhudson (commented):
I'm just lobbing things in from the peanut gallery so I'm probably missing stuff but I have a few comments.

In general there is something slightly ... amusing? ... about the efforts all this goes to to get the grub and shim from the matching series when we are moving to a world where all series have the same grub and shim packages. Did you consider getting the files from the cd-boot-images-$ARCH package?

snapcraft.yaml
lool: Thanks! I've just pushed a commit to fix this; I don't know how to fully test the snap part of this branch, so I might have missed other issues there

Makefile
lool: Ah I did think about omitting -security because it's pocket-copied into -updates regularly, but I should at the very least have a comment explaining this; I've added it now in a new commit, do you think I should always list -security to be on the safe side?

Makefile
lool: Correct; which is currently what happens if you're building a lunar/mantic image from a jammy host using ubuntu-images definitions: ubuntu-image pulls the gadget repo, runs make SERIES=lunar ARCH=arm64, but SERIES was being completely ignored and you ended up with jammy's grub binaries in your lunar image.

Makefile
sil2100: I would like to have -security always included. It's per convention.

sil2100 (approved):
The gadget stuff looks good, chdist parts as well. I didn't build the gadget yet myself though, but at least code-wise this looks sane.

sil2100:
Regarding Michael's comment about cd-boot-images-$ARCH - I *personally* would prefer not to add another dependency on those, although that is an option. I already consider it a bit error-prone and suboptimal that we need to update those packages everytime we have some grub/shim/whatever update and also, I actually don't know what's the future of those? Do we intend to keep those around even if we have newer build tools? I think this needs a bit of discussion.

lool:
I've pushed a commit to always use -security and gave this a local "make" successfully

vorlonofportland:
> 

I agree that we don't want to add a dependency on the cd-boot-images-* indirection.

As far as the future of those packages, we're far enough away today from ubuntu-image being able to build hybrid ISOs with package pools that I wouldn't worry about it.

sil2100 (approved):

sil2100 (approved):

xnox:
> Did you consider getting the files from the cd-boot-images-$ARCH package?

the only purpose of these packages was to:
1. kill d-i builds
2. which ran things on the native arch to generate things that only native arch can do
3. which then are vendored by debian-cd code which doesn't have ability to exeucte foreign arch things

by arch things

cd-boot-images-amd64:
* i386-pc/boot_hybrid.img (can be build during src:grub2 builds but isn't currently)
* msdos efi.img (contains grub.efi and shim.efi, could be built by shim-signed with access to previously signed shim)

If we stop supporting legacy ISO boot, and simply build GPT image with ESP, immutable squashfs or immutable ext4 as the current content of ISO filesystem we don't need any of the above. 

If we want to keep EFI ISO boot, we could figure out how to create msdos efi.img on each build, but it is fiddlly and it is nice to reuse it across image builds.

cd-boot-images-arm64:
* msdos efi.img

as above

cd-boot-images-ppc64el:
* ppc/bootinfo.txt
* grub/powerpc.elf

Both could be generated by builds of src:grub2

PR #101, opened 2023-07-07
Set min-size for amd64 and remove mbr ids for arm64

alfonsosanchezbeato:


mvo5 (approved):
Thank you, this looks fine

PR #106, opened 2023-09-25
22 Support shim fallback bootchain

olivercalder:
This PR adds fallback shim and grub EFI assets to `EFI/ubuntu/`, while leaving non-fallback assets in `EFI/BOOT/`.

This is part of larger work on the snapd side to explicitly set EFI boot variables on install, thus no longer relying on default EFI boot behavior, which can be affected by such things as attaching external media.

The corresponding snapd PR is:
- https://github.com/snapcore/snapd/pull/13205

For those internal to Canonical, the spec related to this work can be found at:
- https://docs.google.com/document/d/1s-XJ56Ur6mQdJ3gVtWdJs12odgknM_2nqQ72RzWUDho

This is a revival of previous PRs which sought to add shim fallback support:
- https://github.com/snapcore/pc-gadget/pull/55
- https://github.com/snapcore/pc-gadget/pull/56

Similar changes should be ported to core24, and perhaps backported to core20 as well (likely not core18 or core16, though).

### DIFF
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -28,9 +28,17 @@ volumes:
           edition: 2
         content:
           - source: grubx64.efi
-            target: EFI/boot/grubx64.efi
+            target: EFI/ubuntu/grubx64.efi
           - source: shim.efi.signed
-            target: EFI/boot/bootx64.efi
+            target: EFI/BOOT/BOOTX64.efi
+          - source: fbx64.efi
+            target: EFI/BOOT/fbx64.efi
+          - source: mmx64.efi
+            target: EFI/BOOT/mmx64.efi
+          - source: BOOTX64.CSV
+            target: EFI/ubuntu/BOOTX64.CSV
+          - source: shim.efi.signed
+            target: EFI/ubuntu/shimx64.efi
       - name: ubuntu-boot
         role: system-boot
         filesystem: ext4
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -25,7 +25,7 @@ volumes:
         type: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
         size: 1200M
         update:
-          edition: 3
+          edition: 2
         content:
           - source: shim.efi.signed
             target: EFI/BOOT/BOOTX64.efi
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -31,8 +31,6 @@ volumes:
             target: EFI/BOOT/BOOTX64.efi
           - source: fbx64.efi
             target: EFI/BOOT/fbx64.efi
-          - source: grubx64.efi
-            target: EFI/BOOT/grubx64.efi
           - source: mmx64.efi
             target: EFI/BOOT/mmx64.efi
           - source: BOOTX64.CSV
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -31,8 +31,6 @@ volumes:
             target: EFI/BOOT/BOOTX64.efi
           - source: fbx64.efi
             target: EFI/BOOT/fbx64.efi
-          - source: mmx64.efi
-            target: EFI/BOOT/mmx64.efi
           - source: BOOTX64.CSV
             target: EFI/ubuntu/BOOTX64.CSV
           - source: grubx64.efi
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -28,6 +28,8 @@ volumes:
           edition: 2
         content:
           - source: grubx64.efi
+            target: EFI/BOOT/grubx64.efi
+          - source: grubx64.efi
             target: EFI/ubuntu/grubx64.efi
           - source: shim.efi.signed
             target: EFI/BOOT/BOOTX64.efi
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -39,6 +39,8 @@ volumes:
             target: EFI/ubuntu/BOOTX64.CSV
           - source: grubx64.efi
             target: EFI/ubuntu/grubx64.efi
+          - source: mmx64.efi
+            target: EFI/ubuntu/mmx64.efi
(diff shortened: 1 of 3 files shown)


gadget/gadget-amd64.yaml
```diff
@@ -25,12 +25,22 @@ volumes:
         type: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
         size: 1200M
         update:
-          edition: 2
+          edition: 3
         content:
           - source: grubx64.efi
-            target: EFI/boot/grubx64.efi
+            target: EFI/BOOT/grubx64.efi
```
valentindavid: This is right.

I was a bit confused on why we still need this. But shim is configured to boot `\grubx64.efi` (`DEFAULT_LOADER`). And `generate_path_from_image_path` from shim considers paths that do not start with `\EFI\` to be relative.

I wonder if it would be better to sort these content entries by target directory. So it is easier to see what goes  in `EFI\BOOT` and what goes in `EFI\ubuntu`.

gadget/gadget-amd64.yaml
```diff
@@ -25,12 +25,22 @@ volumes:
...
+          - source: grubx64.efi
+            target: EFI/ubuntu/grubx64.efi
+          - source: shim.efi.signed
+            target: EFI/BOOT/BOOTX64.efi
+          - source: fbx64.efi
+            target: EFI/BOOT/fbx64.efi
+          - source: mmx64.efi
```
valentindavid: I think the mok manager should also go in `EFI\ubuntu`.

gadget/gadget-amd64.yaml
olivercalder: Thanks, good to know. I totally agree about sorting, I'll fix that in a moment.

gadget/gadget-amd64.yaml
olivercalder: Sounds good. I don't really know what the mok manager is used for, and there was some consternation about adding it in the past: https://github.com/snapcore/pc-gadget/pull/56#discussion_r815868895

It's not one of the boot assets added by the snapd PR #13205 (https://github.com/snapcore/snapd/pull/13205/files#diff-dca334c34d6acf0ce9fb49a95fb836caafe064f7fd2eda43a2185d6d9c7e20e5R464-R476), so I'm not sure if it's something we actually need to add.

What do you think?

mwhudson:
This doesn't build because fb$arch.efi and mm$arch.efi get deleted in snapcraft.yaml, is that expected?

gadget/gadget-amd64.yaml
valentindavid: I would love to have it. The mok keys can be used to sign your own kernels. But it can also be used by anything that uses the secondary trusted keyring in the kernel: kernel modules, signed dm-verity, etc.

Also there is no security argument here, because there is a signed efi that anyone can install manually. We are just making it more annoying to install it for legitimate users.

If it causes some issues and would delay this PR, you do not have to add it.

But for sure, if we add it in `EFI\BOOT`, it should also go in `EFI\ubuntu`. It feels very weird to have it in only one. That would be, you can install machine owner keys, but only if in the bios, you force to boot the default efi bootloader (`EFI\BOOT\bootx64.efi`) instead of the configured `BootOrder`.

To sum up. Either add `EFI\ubuntu\mmx64.efi` or remove `EFI\BOOT\mmx64.efi`.

gadget/gadget-amd64.yaml
```diff
@@ -25,12 +25,22 @@ volumes:
         type: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
         size: 1200M
         update:
-          edition: 2
+          edition: 3
```
alfonsosanchezbeato: Please do not update the edition yet - probably we do not want this change for units in the field, but only apply it to images built from scratch.

gadget/gadget-amd64.yaml
```diff
@@ -25,12 +25,22 @@ volumes:
...
         content:
+          - source: shim.efi.signed
+            target: EFI/BOOT/BOOTX64.efi
+          - source: fbx64.efi
+            target: EFI/BOOT/fbx64.efi
           - source: grubx64.efi
-            target: EFI/boot/grubx64.efi
+            target: EFI/BOOT/grubx64.efi
```
alfonsosanchezbeato: Are we sure we need grub in `EFI/BOOT`? Afaiu reading shim code, shim would never run it if it finds `EFI/BOOT/fbx64.efi`

alfonsosanchezbeato (commented):
This should be proposed to branch 24 first as it is quite a risky change. There should be also an `assumes` in `snapcraft.yaml` for the snapd version that will contain the changes needed for this to happen. Before that this PR cannot be merged.

gadget/gadget-amd64.yaml
olivercalder: @alfonsosanchezbeato do you think `grub${arch}.efi` is required in `EFI/BOOT/` because of this?

alfonsosanchezbeato: Afaict if the fallback binary is in `EFI\boot`, it will get run instead of grub, and even if it does not find any CSV file, `grub*.efi` will not be invoked, so I don't see it necessary. @valentindavid please correct if I am getting something wrong here.

gadget/gadget-amd64.yaml
valentindavid: I think you are right.

gadget/gadget-amd64.yaml
valentindavid: You are right. This is how it seems documented.

gadget/gadget-amd64.yaml
alfonsosanchezbeato: Probably we can live with MokManager only in `EFI\ubuntu`, as we have seen that grub will not ever run from `EFI\boot` if we have fbx64.efi. As MM is called either when shim tries to run grub but it has not a valid signature or when the OS modifies EFI variables related to the MoK - and the OS is expected to have been started by grub in `EFI\ubuntu`.

valentindavid: Sounds correct.

olivercalder: Makes sense. I've removed it now, thanks!

PR #107, opened 2023-10-05
24 Support shim fallback bootchain

olivercalder:
Adjust the EFI boot assets to place shim and grub binaries in `EFI/ubuntu/`, with boot and fallback binaries in `EFI/BOOT/`. Additionally, `BOOT${arch}.CSV` and `mm${arch}.efi` are placed in `EFI/ubuntu/`.

This corresponds to work being done in snapd to support shim fallback: https://github.com/snapcore/snapd/pull/13205

This corresponds to PR #106 , and is designed to be eventually backported to branches 22 and 20.

The possible boot patterns are:
1. If EFI boot variables are set correctly, boot the desired asset correctly (namely, `EFI/ubuntu/shim${arch}.efi`).
2. If the system supports fallback, then `EFI/BOOT/fb${arch}.efi` is run, which searches for `BOOT${arch}.CSV`, finds it in `EFI/ubuntu/`, and thus boots the corresponding assets in `EFI/ubuntu/`.
3. If fallback is not supported, the system directly boots `EFI/BOOT/BOOT${arch}.efi`.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -25,9 +25,9 @@ hooks:
       # DO NOT check this API key into a publicly accessible VCS
       MODEL_APIKEY: ""
 
-# Min version to support shim 15.7
+# Min version to support shim 15.7 and assets in /EFI/ubuntu
 assumes:
-  - snapd2.59.3
+  - snapd2.61
 
 parts:
   mbr:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -52,27 +52,44 @@ parts:
       grub_target=x86_64
       grub_bin=grubx64.efi.signed
       shim_bin=shimx64.efi.dualsigned
+      fb_bin=fbx64.efi
+      mm_bin=mmx64.efi
+      boot_csv=BOOTX64.CSV
       if [ "$CRAFT_TARGET_ARCH" = arm64 ]; then
           grub_target=arm64
           grub_bin=grubaa64.efi.signed
           shim_bin=shimaa64.efi.dualsigned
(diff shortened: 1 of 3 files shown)


snapcraft.yaml
```diff
@@ -52,27 +52,44 @@ parts:
...
+      #sbverify --list "$fb_path"   | grep -E 'Canonical Ltd. Secure Boot Signing \(Ubuntu Core'
+      #sbverify --list "$mm_path"   | grep -E 'Canonical Ltd. Secure Boot Signing \(Ubuntu Core'
 
-      # Move shim/grub to the expected path
+      # Move assets to the expected paths
       install -m 644 "$shim_path" "$CRAFT_PART_INSTALL"/shim.efi.signed
       install -m 644 "$grub_path" "$CRAFT_PART_INSTALL"/${grub_bin%.signed}
+      install -m 644 "$fb_path"   "$CRAFT_PART_INSTALL"/$fb_bin
+      install -m 644 "$mm_path"   "$CRAFT_PART_INSTALL"/$mm_bin
+      install -m 644 "$csv_path"  "$CRAFT_PART_INSTALL"/$boot_csv
```
alfonsosanchezbeato, line 83: Maybe time to make this a for loop

alfonsosanchezbeato (commented):
Looks great, some comments though:
- Looks like the right time to remove shim from the boot partition (it has never been used)
- 2.61 might be too optimistic
See also comment below.

In any case, the PR cannot be merged until a snapd version with support for this has landed.

olivercalder:
This PR is being taken over by @valentindavid

PR #108, opened 2023-10-05
Bump edition for assets

alfonsosanchezbeato:


sil2100 (approved):
Ah, thanks for demonstrating how this work! Let's get it merged and use this for the boot asset upgrade test.

PR #111, opened 2024-02-14
prepare-device: implement reference prepare-device hook

tonyespy:
Snapcraft.io points to this as the "reference" implementation of the amd64 gadget. However, no prepare-device hook makes this an incomplete reference, especially for private store operators

The proposed prepare-device hook would not change the behavior of the default PC gadget. However, it could then be forked and trivially modified to use for private store authentication.

This PR is just a backport of the reference hook added to the `22` and `24` branches.

### DIFF
--- a/snap/hooks/prepare-device
+++ b/snap/hooks/prepare-device
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+# This file is included for reference purposes. If MODEL_APIKEY is not set
+# in snapcraft.yaml (as is the default), it is a no-op. For more about how
+# to use this script to connect a device to an IoT App Store, see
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
Tbh I'm a bit reluctant to backport this sort of changes, as this is just for reference and it is in the more modern gadgets. And new projects should be based on those, not on 20.

xnox (approved):
Given how confusing all of this is, I'd rather have this here, than not.

xnox:
also can we just make MODEL_APIKEY somehow snapctl set'able? such that verbantim pc gadget can just be used with serial vaults without need to maintain a forked gadged?

PR #113, opened 2024-04-04
gadget: use new boot chain

valentindavid:
pc-gadget provides the shim boot chain for removable media. This is not the proper boot chain for an installed system.
    
The difference between boot chains is explained upstream: https://github.com/rhboot/shim/blob/main/README.fallback
    
This does not introduce the MOK manager yet.
    
This depends on snapd 2.63 because downgrade would require removing `EFI/boot/fbx64.efi`.

Depends on:
* https://github.com/snapcore/snapd/pull/13412
* https://github.com/snapcore/snapd/pull/13734

### DIFF
--- a/gadget/gadget-amd64.yaml
+++ b/gadget/gadget-amd64.yaml
@@ -32,7 +32,13 @@ volumes:
           edition: 2
         content:
           - source: grubx64.efi
-            target: EFI/boot/grubx64.efi
+            target: EFI/ubuntu/grubx64.efi
+          - source: shim.efi.signed
+            target: EFI/ubuntu/shimx64.efi
+          - source: boot.csv
+            target: EFI/ubuntu/bootx64.csv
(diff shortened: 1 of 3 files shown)


alfonsosanchezbeato:
Is this equivalent to https://github.com/snapcore/pc-gadget/pull/107 , I assume?

valentindavid:
> Is this equivalent to #107 , I assume?

Yes.

valentindavid:
Though the difference is this does not add the mokmanager.

alfonsosanchezbeato (approved):
LGTM, thanks!

alfonsosanchezbeato (commented):
But, I think we need a "assumes" on the version of snapd that supports this before merging.

valentindavid:
> But, I think we need a "assumes" on the version of snapd that supports this before merging.

It should already be there because of #116

PR #115, opened 2024-04-10
snapcraft.yaml: downgrade shim

valentindavid:
New shim uses a new sbatlevel that is not supported yet by snapd.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -42,9 +42,34 @@ parts:
     plugin: nil
     build-packages:
       - sbsigntool
+      - wget
+      - attr
     stage-packages:
       - grub-efi-$CRAFT_ARCH_BUILD_FOR-signed
-      - shim-signed
+    build-environment:
+      - shim_signed_version: 1.56+15.7-0ubuntu1
+    override-pull: |
+      craftctl default
+
+      # TODO: get shim as a stage package and add snapd 2.63 as requirement
+      case "${CRAFT_ARCH_BUILD_FOR}" in
+        amd64)
+          wget "http://archive.ubuntu.com/ubuntu/pool/main/s/shim-signed/shim-signed_${shim_signed_version}_amd64.deb"
+          sha256sum -c <<EOF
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -59,15 +83,18 @@ parts:
...
       # Make sure we have have the right signatures
-      shim_path="${CRAFT_PART_INSTALL}"/usr/lib/shim/$shim_bin
+      shim_path="shim-signed_${shim_signed_version}/usr/lib/shim/$shim_bin"
       grub_path="${CRAFT_PART_INSTALL}"/usr/lib/grub/"$grub_target"-efi-signed/$grub_bin
       sbverify --list "$shim_path" | grep -E 'Canonical Ltd. Secure Boot Signing \(2022 v1\)'
       sbverify --list "$grub_path" | grep -E 'Canonical Ltd. Secure Boot Signing \(2022 v1\)'
 
       # Move shim/grub to the expected path
-      install -m 644 "$shim_path" "${CRAFT_PART_INSTALL}"/shim.efi.signed
+      cp -a "$shim_path" "${CRAFT_PART_INSTALL}"/shim.efi.signed
```
alfonsosanchezbeato, line 93: Why do you prefer cp here?

snapcraft.yaml
```diff
@@ -42,9 +42,33 @@ parts:
     plugin: nil
     build-packages:
       - sbsigntool
+      - wget
+      - attr
     stage-packages:
       - grub-efi-$CRAFT_ARCH_BUILD_FOR-signed
-      - shim-signed
```
alfonsosanchezbeato, line 47: Maybe comment instead of removing and add a TODO to revert things when 2.63 is out

alfonsosanchezbeato (commented):
Thanks! see comment and question below

snapcraft.yaml
valentindavid: To keep the extended attributes so that snapcraft finds the original package.

snapcraft.yaml
valentindavid: I have added a TODO in the override-pull

alfonsosanchezbeato (approved):
LGTM, thanks

PR #116, opened 2024-04-17
Revert #115

valentindavid:
We can revert #115 as soon as snapd 2.63, as it should contain fix https://github.com/snapcore/snapd/pull/13807

We should not merge until we have snapd 2.63 actually available in latest/stable.


Meulengracht (approved):
nice, LGTM

alfonsosanchezbeato (approved):
LGTM, thanks

PR #118, opened 2024-05-03
Define a system-data volume with 5GB size (LP#2064322)

rmartin013:
ubuntu-images fails to create valid GPT partition table (the backup GPT header isn't located at the right position and also invalid) when the gadget snap doesn't include an explicit system-data structure.
Defining one here with 5G size (that suits the 4,8GB of rootfs partition for ubuntu-tegra classic server images), and using the "writable" name/label to keep the same image attribute.

### DIFF
--- a/gadget-arm64.yaml
+++ b/gadget-arm64.yaml
@@ -14,3 +14,9 @@ volumes:
             target: EFI/boot/bootaa64.efi
           - source: grub.cfg
             target: EFI/ubuntu/grub.cfg
+      - name: writable
+        role: system-data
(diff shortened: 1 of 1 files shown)


upils:
@waveform80 @sil2100 would you mind taking a look? This is in the context of [LP #2064322](https://bugs.launchpad.net/ubuntu-image/+bug/2064322)

sil2100 (approved):
As a temporary solution this feels fine. But let's resolve this properly and switch back to the classic gadget when possible.

PR #126, opened 2024-10-03
set up an oracular chdist instead of using stage-packages  ￼

mwhudson:
Now lightly tested.


sil2100 (commented):
Looks like this could work, but the snap still fails to build for now!

mwhudson:
it builds now!

sil2100 (approved):
Ok, this is certainly not the most beautiful thing I ever saw, but it seems to work. I checked the resulting snap and we seem to have pulled in the oracular bootloader bits. It's certainly an improvement. Let's try this on LP.

PR #140, opened 2025-12-15
pc-gadget: document OpenSSL-based secure onboarding hooks (commented reference)

TSainathPolisetty:
## Summary

This PR adds **fully commented, non-executing reference implementations**
for `prepare-device` and `prepare-serial-request` hooks illustrating an
**OpenSSL-based secure onboarding flow** for Ubuntu Core devices.

All logic is intentionally commented out. There is **no functional change**
to the existing gadget behavior.

## Motivation

This is meant to serve as:
- A **readable reference** for implementing secure onboarding
- A concrete mapping between the **secure onboarding specification** and
  hook-level implementation details
- A starting point for future work without changing current defaults

## Design notes

- Uses OpenSSL for RSA-2048 key generation and SHA256 signing
- Private key is stored in `SNAP_SAVE_DATA` and generated once per device
- `prepare-device` handles identity material only
- `prepare-serial-request` demonstrates request-id signing and payload construction
- Network operations and snapd defaults are untouched

## Scope / Safety

- All code is commented out
- No hooks are executed
- No runtime behavior is modified
- No serial service is contacted

## Follow-ups

If useful, this reference can later be:
- Moved to documentation
- Gated behind a build option
- Used as a basis for a TPM-backed variant


mcw-work:
This PR probably obsoletes PR #111 but does reflect the greater functionality we now have for device registration.

